PR #1, opened 2019-01-07
 Fix environments:copy-layout command 

ArnaudValensi:


test/commands/environments/copy-layout.test.js
```diff
@@ -1,11 +1,18 @@
-/* global describe */
```
ArnaudValensi: Useless with the config in `test/.eslintrc`

ArnaudValensi:
[(0.5) Copy Layout - Users want to have a `jobId` in response `meta` to be able to know when the job is done.](https://trello.com/c/vSJzP7BH/3097-05-copy-layout-users-want-to-have-a-jobid-in-response-meta-to-be-able-to-know-when-the-job-is-done)

src/serializers/deployment-request.js
```diff
@@ -1,14 +1,6 @@
...
-    ref: 'id',
-    included: false,
-  },
-  to: {
-    ref: 'id',
-    included: false,
-  },
-  typeForAttribute: () => 'environments',
+  attributes: ['type', 'from', 'to'],
+  typeForAttribute: () => 'deployment-requests',
```
VincentMolinie: This is useless. There is no attribute with not foreignKey so typeForAttribute will never be called. It can be removed

test/commands/environments/copy-layout.test.js
```diff
@@ -1,11 +1,18 @@
...
+const { expect } = chai;
+chai.use(chaiSubset);
 
 const fancy = test.register('nock', Nock);
 const EnvironmentSerializer = require('../../../src/serializers/environment');
 
 describe('environments:copy-layout', () => {
+  let parsedBody;
+
   const mocks = fancy
```
VincentMolinie: Why is this not in a before ?

VincentMolinie:
All good after that :rocket:

test/commands/environments/copy-layout.test.js
ArnaudValensi: David modified the tests in the next PR

PR #8, opened 2019-02-12
[*] Authentication - If the user session has perished or user corrupted the token where it is store we should automatically logout the user

ArnaudValensi:


ArnaudValensi:
[(0.5) Toolbelt - If the user session has perished on the toolbelt (15 days have past, or user corrupted the token where it is stored,...) we should automatically logout the user from the toolbelt.](https://trello.com/c/WuYeoBXj/3211-05-toolbelt-if-the-user-session-has-perished-on-the-toolbelt-15-days-have-past-or-user-corrupted-the-token-where-it-is-stored-we)

PR #10, opened 2019-03-15
[+] Technical - Warn users if an update of the Toolbelt is available

root-io:


root-io:
[Lumber/Toolbelt - Warn user if an update is available.](https://trello.com/c/kHzTVkta/3294-lumber-toolbelt-warn-user-if-an-update-is-available)

PR #17, opened 2020-01-06
test: use jest instead of mocha for the test base

arnaudbesnier:


.eslintrc.js
```diff
@@ -2,13 +2,27 @@ module.exports = {
...
+        devDependencies: [
+          'scripts/*.js',
+          'test/**/*.js'
+        ]
+      },
+    ],
+    'jest/no-hooks': ['error', {
+      allow: ['beforeEach'],
+    }],
+    'jest/no-standalone-expect': 0,
```
arnaudbesnier: Needs to be deactivated to work with oclif tests.

.eslintrc.js
```diff
@@ -2,13 +2,27 @@ module.exports = {
...
+      'error',
+      {
+        devDependencies: [
+          'scripts/*.js',
+          'test/**/*.js'
+        ]
+      },
+    ],
+    'jest/no-hooks': ['error', {
+      allow: ['beforeEach'],
```
arnaudbesnier: Needs to be deactivated to prevent authentication context issues in some test files.

.eslintrc.js
```diff
@@ -2,13 +2,27 @@ module.exports = {
...
     'implicit-arrow-linebreak': 0,
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        devDependencies: [
+          'scripts/*.js',
+          'test/**/*.js'
+        ]
+      },
+    ],
```
arnaudbesnier: Add a missing basic rule to enable required dev-dependencies in scripts and tests directories

slimee:
I think we should merge this state and continue to migrate tests from time to time.
 Wdyt?

slimee:
> We should merge this.
> The state of this PR is:
> -rebased,
> -build and pass test and lint (at least as configured),
> -Contain many test migrations as they should be written now.
> 
> We can then:
> -write new test style,
> -migrate old ones from time to time. (very quick now 👍 )
> -unrelove lint rules from time to time.
> 
> 🚀 ? @arnaudbesnier

It must be rebased again.

slimee:
It's rebased, there is now only 3 test files that still use the old 'oclif/test' way of testing.

slimee:
get rid of @oclif/test, which was not living.
All tests use now testCli method.

PR #20, opened 2020-03-03
[+] Versioning - Configure the `warn-if-update-available` plugin that warns when a new version is available

slimee:
… warns when a new version is available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,13 @@
     "plugins": [
       "@oclif/plugin-help",
       "@oclif/plugin-warn-if-update-available"
-    ]
+    ],
+    "warn-if-update-available": {
+      "timeoutInDays": 7,
+      "message": "<%= config.name %> update available from <%= chalk.greenBright(config.version) %> to <%= chalk.greenBright(latest) %>.",
+      "registry": "https://www.npmjs.com/package/forest-cli",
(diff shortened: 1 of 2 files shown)


slimee:
This one is quite hard to test 🤕

package.json
```diff
@@ -72,7 +72,13 @@
       "@oclif/plugin-help",
       "@oclif/plugin-warn-if-update-available",
       "@oclif/plugin-not-found"
-    ]
+    ],
+    "warn-if-update-available": {
+      "timeoutInDays": 7,
+      "message": "<%= config.name %> update available from <%= chalk.greenBright(config.version) %> to <%= chalk.greenBright(latest) %>.",
+      "registry": "https://www.npmjs.com/package/forest-cli",
+      "authorization": "Basic f9feb3a6-a76b-4bd0-ab2f-471e6e46cc56"
```
rap2hpoutre: According to this page (https://github.com/oclif/plugin-warn-if-update-available#configuration) authorization is required for registries that require auth (as far as I know https://www.npmjs.com/package/forest-cli is public).

> authorization - Authorization header value for registries that require auth.

rap2hpoutre: Also I'm not sure about this token: does it allow to read all our private repositories?

I guess we should avoid sharing it!

slimee: You are right, we can remove the auth.

PR #25, opened 2020-03-10
[+] Ask credentials when a command requires it.

slimee:
This PR ask for credentials when needed.
It's needed when the user performs a login-required command.

Note: All commands require login except `forest login` and `forest user`.

### DIFF
--- a/src/services/authenticator.js
+++ b/src/services/authenticator.js
@@ -7,6 +7,7 @@ const jwtDecode = require('jwt-decode');
 const Joi = require('joi');
 const api = require('./api');
 const logger = require('./logger');
+const ERROR_UNEXPECTED = require('../utils/messages');
 
 function Authenticator() {
   this.getAuthToken = (path = os.homedir()) => {
@@ -75,10 +76,18 @@ function Authenticator() {
     });
   };
 
-  this.loginWithEmailOrTokenArgv = async (config) => {
-    const token = await this.login(config);
-    this.saveToken(token);
-    return null;
+  this.tryLogin = async (config) => {
+    await this.logout({ log: false });
+    try {
+      const token = await this.login(config);
+      this.saveToken(token);
+      logger.info('Login successful');
+    } catch (error) {
+      const message = error.message === 'Unauthorized'
+        ? 'Incorrect email or password.'
+        : `${ERROR_UNEXPECTED} ${chalk.red(error)}`;
+      logger.error(message);
+    }
   };
 
   this.loginWithToken = (token) => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Changed
 - Technical - Clean temporary tests files.
+- Login command - Perform login before any authenticated operation if not already logged.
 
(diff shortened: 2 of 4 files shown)


src/services/authenticator.js
```diff
@@ -75,10 +76,18 @@ function Authenticator() {
...
+    await this.logout({ log: false });
+    try {
+      const token = await this.login(config);
+      this.saveToken(token);
+      logger.info('Login successful');
+    } catch (error) {
+      const message = error.message === 'Unauthorized'
+        ? 'Incorrect email or password.'
+        : `${ERROR_UNEXPECTED} ${chalk.red(error)}`;
+      logger.error(message);
```
SteveBunlon: [Thinkin] We could maybe ask again for username / password if the user missed. Just a though, this is ignorable

src/services/authenticator.js
```diff
@@ -75,10 +76,18 @@ function Authenticator() {
     });
   };
 
-  this.loginWithEmailOrTokenArgv = async (config) => {
-    const token = await this.login(config);
-    this.saveToken(token);
-    return null;
+  this.tryLogin = async (config) => {
```
SteveBunlon: [Improvement] This doesn't return anything. We could probably return `true`/`false` to have an instant feedback about the success of the login in the code.

rap2hpoutre: I understand your point. In this case it's a procedure (i.e. not a function) with an imperative name (and behavior) so I guess it execute a sequence of orders. In other word, as a user of this procedure, I'm not sure what a `trySomething` return value would mean (but I would know what a `isSomething` will return a boolean and have no side effect). **Not sure about my opinion though!**

rap2hpoutre: I understand your point. In this case it's a procedure (i.e. not a function) with an imperative name (and behavior) so I guess it execute a sequence of orders. In other word, as a user of this procedure, I'm not sure what a `trySomething` return value would mean (but I would know what a `isSomething` will return a boolean and have no side effect). **Not sure about my opinion though!**

src/services/authenticator.js
slimee: I understand. 
The other hand is when the command ends, we can test it.
(Today we do not know how to test dialog that never end)

slimee:
> Great! It works.
> 
> Maybe we could consider announcing something like `Please login before running this action` (my text as to be challenged, I don't like it!) before prompting for login. Because that's actually not clear when we ask for email if we don't remember that we were logged out.
> 
> Capture d’écran 2020-03-18 à 10 55 29

I really aggree with this 👍

src/services/authenticator.js
slimee: the result of this method is: throw something or success.
I can set 'return true' instead of 'return null', but there is no significance.
I would prefer to return nothing but lint does not want it :)

rap2hpoutre:
(I fixed the "missing semicolon" error [here](https://github.com/ForestAdmin/toolbelt/pull/25/commits/68cb0d7d8de1cbb7ed3d3445a1850adb8275ad97)).

PR #26, opened 2020-03-10
Add the environments:update command

slimee:


slimee:
> According to the [clickup note](https://app.clickup.com/t/3mgfhx):
> 
> > This command looks unusable since the users will never have their environment id to use this command.
> > Wait to discus with owners.
> 
> So this command is still "unusable", right?

It is usable, yes.
I had the opinion that users will not possess their environment ID to use this command.
Without any feedback about this opinion, I just do the work :)

slimee:
> Cool, it works (see my comment about the "usability")!
> 
> Maybe we could suggest that the `name` and `url` are params that we actually want to update (I did not understood at first glance, I thought: "oh, it's required to match the environment, but what do I update?").

I understand that. I asked this to myself. 
An update must specify which is the functionnal key (the WHERE)
And the updatable fields (the SET)

PR #27, opened 2020-03-12
[-] Technical/authentication tests with mock input

slimee:
This PR just adds the "google auth" test.
It updates the API file in order to lazy-evaluate values to allow test mocking. 

It introduces the testCli method.


test/commands/login.test.js
```diff
@@ -1,38 +1,108 @@
...
+        ],
+      }));
+    });
+    describe('with a google mail', () => {
+      describe('with a valid token from input', () => {
+        it('should login successful', () => testCli({
+          env: { FOREST_URL: 'http://localhost:3001' },
+          nock: nock('http://localhost:3001')
+            .get('/api/users/google/robert@gmail.com')
+            .reply(200, { data: { isGoogleAccount: true } }),
```
SteveBunlon: [Thinking] This looks like test environment configuration. We could maybe isolate it somewhere, as this is done with `mirage` in the front app. 

We would have one mockable server, activable or not, on which we could mock some calls.

This is just a thought and is ignorable.

test/commands/test-cli.js
```diff
@@ -0,0 +1,39 @@
+const mockStdin = require('mock-stdin');
+const { stdout, stderr } = require('stdout-stderr');
+const { expect } = require('chai');
+
+module.exports = async function testCli({
+  nock, env, inputs, command, outputs, errorOutputs,
+}) {
+  const previousEnv = process.env;
+  if (env) process.env = env;
```
SteveBunlon: I don't think this `test-cli` file should be responsible of the env. This should be done in the test file, which should describe its own context and restore it right after the tests execution.

test/commands/login.test.js
```diff
@@ -1,38 +1,103 @@
...
+      }));
+    });
+    describe('with valid token in args', () => {
+      const token = jwt.sign({}, 'key', { expiresIn: '1day' });
+      it('should login successful', () => testDialog({
+        command: () => LoginCommand.run(['-e', 'smile@gmail.com', '-t', token]),
+        dialog: [
+          { in: `${jwt.sign({}, 'key', { expiresIn: '1day' })}` },
+          { out: 'Login successful' },
+        ],
```
SteveBunlon: [Thinking] By using this dialog we test two things there according to me. We test the result of the command, and what the command displays in the console. I'm not sure we should tie this two tests together. I have the feeling it will be hard to maintain / subject to bad dev experience.

src/services/api.js
```diff
@@ -7,11 +7,11 @@ const HEADER_FOREST_ORIGIN = 'forest-origin';
 const HEADER_USER_AGENT = 'User-Agent';
 
 function API() {
-  this.endpoint = process.env.FOREST_URL || 'https://api.forestadmin.com';
+  this.endpoint = () => process.env.FOREST_URL || 'https://api.forestadmin.com';
```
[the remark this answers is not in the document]
slimee: It's in function to be lazily evaluated.
This is needed by tests, because process.env is modified after files requires.

test/commands/login.test.js
slimee: I agree. Let's build this together :)

test/commands/test-cli.js
slimee: It's the case, env come from the tests.

test/commands/login.test.js
slimee: Ok. we will have the occasion to improve this.

PR #36, opened 2020-03-20
[+] Tests - Add test for `forest environments:create` command

slimee:
Only tests files are changed here.
One test added, indentation changed for 2 tests.


slimee:
> 🥳

Thank you for these emojis that always give me light in my day :)

PR #37, opened 2020-03-20
[*] Logout - Display an explicit message when lumber is still connected.

slimee:
Few changes are made in order to be simple.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Login command - Perform login before any authenticated operation if not already logged.
 - Commands - Ease the selection of project.
 - Technical - Remove caching of authToken and api endpoint.
-- Logout - Display an explicit message when lumber is still connected.
+- Logout command - Display an explicit message when users having a lumber session tries to logout.
 
 ## RELEASE 1.0.5 - 2020-03-10
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Login command - Perform login before any authenticated operation if not already logged.
 - Commands - Ease the selection of project.
 - Technical - Remove caching of authToken and api endpoint.
-- Logout command - Display an explicit message when users having a lumber session tries to logout.
+- Logout command - Display an explicit message when users having a lumber session try to logout.
 
 ## RELEASE 1.0.5 - 2020-03-10
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Login command - Perform login before any authenticated operation if not already logged.
 - Commands - Ease the selection of project.
 - Technical - Remove caching of authToken and api endpoint.
+- Logout - Display an explicit message when lumber is still connected.
 
 ## RELEASE 1.0.5 - 2020-03-10
 ### Fixed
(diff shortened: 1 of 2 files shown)


src/services/authenticator.js
```diff
@@ -51,29 +50,17 @@ function Authenticator() {
...
-        } else {
-          reject(err);
-        }
-      });
-    });
+    const forestrcPath = `${os.homedir()}/.forestrc`;
+    const isForestLoggedIn = this.getVerifiedToken(forestrcPath);
+    const lumbercPath = `${os.homedir()}/.lumberrc`;
+    const isLumberLoggedIn = this.getVerifiedToken(lumbercPath);
+
```
arnaudbesnier: [Readability] I prefer to start variable names by what kind of object they represent and then add adjectives.
```suggestion
    const pathForestrc = `${os.homedir()}/.forestrc`;
    const pathLumberc = `${os.homedir()}/.lumberrc`;
    const isLoggedInWithForestCLI = this.getVerifiedToken(forestrcPath);
    const isLoggedInWithLumberCLI = this.getVerifiedToken(lumbercPath);

```

src/services/authenticator.js
```diff
@@ -51,29 +50,17 @@ function Authenticator() {
...
+    if (isForestLoggedIn) {
+      fs.unlinkSync(forestrcPath);
+    }
+    if (opts.log) {
+      logger.info(`You are unlogged. ${isLumberLoggedIn ? '(still connected via lumber)' : ''}`);
```
arnaudbesnier: "(still connected via lumber)" => so Forest CLI will behave as if you were logged.
if really unlogged:
`console.log(chalk.green('You are now unlogged.'));`
(Why do you remove the color by the way? info is green?)

if logged in with lumber you should display a very specific message:
"You cannot be logged out with this command. Please use lumber logout" (something like that).

src/services/authenticator.js
```diff
@@ -51,29 +50,17 @@ function Authenticator() {
     || 'Invalid token. Please enter your authentication token.';
 
   this.logout = async (opts = {}) => {
-    const path = `${os.homedir()}/.forestrc`;
-
-    return new P((resolve, reject) => {
-      fs.stat(path, (err) => {
```
arnaudbesnier: Out of curiosity, why has it been removed?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Login command - Perform login before any authenticated operation if not already logged.
 - Commands - Ease the selection of project.
 - Technical - Remove caching of authToken and api endpoint.
+- Logout - Display an explicit message when lumber is still connected.
```
arnaudbesnier:
```suggestion
- Logout command - Display an explicit message when a users having a lumber session tries to logout.
```

src/services/authenticator.js
slimee: it was doing the job 'isLoggedIn?' by just checking the presence of the `.forestrc` file.
Now we rely on the `getVerifiedToken`because:
-it's same code than we use to evaluate at login-in
-it's smarter (validate the token) 
-it's more simple (no promise)

src/services/authenticator.js
slimee: about green color: yes, logger is preferred over console, it manage the console. I've set it to green, let me know if it's not better.

PR #39, opened 2020-03-21
[+] Technical - Add sonar lint

slimee:


slimee:
> We could have created two PRs: one for the prompter refactoring and the other one for sonar. WDYT?
> 
> (also we could consider adding an entry in changelog)

Thanks for your review!
I've update the changelog.
By the way, I've considered the package very simple, and on the same subject to be in one PR.
Let me know if you consider separating the 'add sonar' stuff still a needed thing. 👍

PR #46, opened 2020-04-24
chore(tests): add unexpected error management in tests

slimee:
This PR allows to a better display in case of unexpected errors.

Given a test that should success (no error thrown)
When the test run but throw an error (unexpected)
Then a friendly message explains the situation (command throw this unexpected error: ...)

To test it:
-Take an existing error test (example l.31 of `delete.test.js`)
-Replace the expected `exitCode` by an expected output like std:[{out:"toto"}]
-The test should fail and display the reason.

### DIFF
--- a/test/commands/test-cli-errors.js
+++ b/test/commands/test-cli-errors.js
@@ -44,8 +44,11 @@ function errorIfBadCommand(command) {
 
 function errorIfNoStd(stds) {
   if (!stds || !Array.isArray(stds) || !stds.length > 0) {
-    throw new Error('testCli configuration error: "std" must be an not empty array ex.'
-      + ' [{in:\'john\'},{out:\'hello, john\'}]');
+    throw new Error('testCli configuration error: "std" or "exitCode" or "exitMessage" must be' +
+      ' defined.\n' +
+      ' "std" must be a not empty array like [{in:\'john\'},{out:\'hello, john\'}]\n' +
+      ' Define "exitCode" and/or "exitMessage" instead of "std" if you are testing an error' +
+      ' case.');
   }
 }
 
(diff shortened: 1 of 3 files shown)


test/commands/test-cli-errors.js
```diff
@@ -44,8 +44,11 @@ function errorIfBadCommand(command) {
 
 function errorIfNoStd(stds) {
   if (!stds || !Array.isArray(stds) || !stds.length > 0) {
-    throw new Error('testCli configuration error: "std" must be an not empty array ex.'
-      + ' [{in:\'john\'},{out:\'hello, john\'}]');
+    throw new Error('testCli configuration error: "std" or "exitCode" or "exitMessage" must be'
+      + ' defined.\n'
+      + ' "std" must be a not empty array like [{in:\'john\'},{out:\'hello, john\'}]\n'
+      + ' Define "exitCode" and/or "exitMessage" instead of "std" if you are testing an error'
+      + ' case.');
```
larcin, line 51: [code style] I think that we tend to avoid string building/concatenation using `+`. You could replace it by backslashes.

test/commands/test-cli-errors.js
```diff
@@ -60,6 +63,16 @@ function validateInput(file, command, stds, expectedExitCode, expectedExitMessag
   }
 }
 
+// NOTICE: Assert that command did not throw an error if there is no expected error.
+function assertNoErrorThrown(actualError, expectedExitCode, expectedExitMessage) {
+  if (expectedExitCode || expectedExitMessage) return;
+
+  const noErrorMessage = 'Unexpected error thrown by command.\n'
+    + ' No "exitCode" and/or "exitMessage" is specified, so this error should not be thrown.';
```
larcin, line 71: [code style] Same here, you can use `\`

jeffladiray:
> Just a couple of comments.
> 
> Plus you have to change the pull request name to make it conventional commit compliant (ie. `chore(tests): add unexpected error management in tests`) => That's what breaks the build

PR name was not enough, it seems that travis also check for the full branch commit message history. I reworded the culprit 😈
`\` concatenation is not authorized by the current lint rules, that's why I added some lint ignore (But I guess since it's only in a test mock file it's ok).
SGTM now, and it will really ease the test writing process for futur PRs.

PR #47, opened 2020-04-27
feat(branch-command): add branch creation

larcin:


src/services/branch-manager.js
```diff
@@ -6,8 +16,53 @@ async function deleteBranch() {
...
+function handleError(error) {
+  try {
+    const { status } = error;
+
+    if (status === 404) {
+      // NOTICE: When no env/project can be found through envSecret
+      return ERROR_MESSAGE_ENV_SECRET_ISSUE;
+    }
+
+    const message = JSON.parse(error.response.text).errors[0].detail;
```
jeffladiray, line 38: [Suggestion] Since we'll have to work on other commands, can't we create an `error` deserializer ?

src/services/branch-manager.js
```diff
@@ -1,3 +1,13 @@
+const P = require('bluebird');
+const agent = require('superagent-promise')(require('superagent'), P);
```
jeffladiray, line 2: [Suggestion] Not mandatory (Since I also add it in my PR), but it would be nice to remove `bluebird` here I guess, since `Promise` should work ? (Didn't test)

src/services/branch-manager.js
```diff
@@ -6,8 +16,53 @@ async function deleteBranch() {
...
 
-async function createBranch() {
-  // FIXME: Implement createBranch function
+async function createBranch(branchName) {
+  const authToken = authenticator.getAuthToken();
+
+  await agent
+    .post(`${serverHost()}/api/branches`)
+    .set('Authorization', `Bearer ${authToken}`)
+    .set('forest-secret-key', `${process.env.FOREST_ENV_SECRET}`)
```
jeffladiray, line 25: Since we'll have to add the `--project` anytime soon, maybe the `envSecret` should be passed to the function instead of using `process.env` here ?

src/commands/branch.js
```diff
@@ -26,12 +26,14 @@ class BranchCommand extends AbstractAuthenticatedCommand {
...
       return this.log(`✅ Switched to new branch: ${branchName}.`);
     } catch (err) {
+      const customError = BranchManager.handleError(err);
       // FIXME: Display correct error, depending on the case
       //        - Branch already exist
       //        - Server branch creation failed
       //        - Absence of a remote or a production environment
       //        Cases: #5, #6, #7
-      return this.error(`❌ Failed create ${branchName}.`);
+      // return this.error(`❌ Failed create ${branchName}.`);
```
jeffladiray, line 35: This could be removed, WDYT ?

src/services/branch-manager.js
```diff
@@ -1,3 +1,13 @@
+const P = require('bluebird');
+const agent = require('superagent-promise')(require('superagent'), P);
+const authenticator = require('./authenticator');
+const { serverHost } = require('../config');
+
+const ERROR_MESSAGE_PROJECT_IN_V1 = '⚠️  This project does not support branches yet. Please migrate your environments from your Project settings first.';
+const ERROR_MESSAGE_ENV_SECRET_ISSUE = '⚠️  Your development environment is not properly set up. Please run `forest init` first and retry.';
+const ERROR_MESSAGE_BRANCH_ALREADY_EXISTS = '❌ This branch already exists.';
+const ERROR_MESSAGE_NO_PRODUCTION_OR_REMOTE_ENVIRONMENT = '❌ You cannot create a branch until this project has either a remote or a production environment.';
```
jeffladiray, line 9: I think this error might also happend for all the commandes, so maybe it should be rephrased (Like `You cannot run branch commands until this project has either a remote or a production environment.`) ?

PR #48, opened 2020-04-27
feat(branch-command): add branch list command

jeffladiray:


src/services/branch-manager.js
```diff
@@ -8,8 +9,14 @@ const ERROR_MESSAGE_ENV_SECRET_ISSUE = '⚠️  Your development environment is
...
-  // FIXME: Implement getBranches function
+async function getBranches(envSecret) {
+  const authToken = authenticator.getAuthToken();
+  return agent
+    .get(`${serverHost()}/api/branches`)
+    .set('Authorization', `Bearer ${authToken}`)
+    .set('forest-secret-key', envSecret)
+    .send()
+    .then((response) => branchDeserializer.deserialize(response.body));
 }
```
larcin, line 20: [Code consistency] Here you have an async function with no await, you can drop it.

Moreover it seems that getBranches returns a Promise, while in createBranch, I'm waiting for the promise to settle.

In order to be consistent all over this file, would you mind editing mine to remove both await and async? (I do prefer your way)

PR #49, opened 2020-04-27
feat(init command): create the init command with the login 1st step (#4wman4)

adriguy:
https://app.clickup.com/t/4wman4

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -16,7 +16,6 @@ InitCommand.flags = {
   project: flags.string({
     char: 'p',
     description: 'The name of the project you want to init.',
-    default: null,
   }),
 };
 
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -1,0 +1,23 @@
+const { flags } = require('@oclif/command');
+const AbstractAuthenticatedCommand = require('../abstract-authenticated-command');
+const logger = require('../services/logger');
+
+class InitCommand extends AbstractAuthenticatedCommand {
+  async runIfAuthenticated() {
+    // TO BE REMOVED. JUST CHECKING IF THE FLAG PARSING IS WORKING HERE :)
+    const parsed = this.parse(InitCommand);
+    logger.info('working here 🤙 flag:', parsed.flags);
+  }
+}
+
+InitCommand.description = 'Set up your development environment in your current folder.';
+
+InitCommand.flags = {
+  project: flags.string({
+    char: 'p',
+    description: 'The name of the project you want to init.',
+    default: null,
+  }),
+};
+
+module.exports = InitCommand;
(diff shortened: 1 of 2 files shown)


src/commands/init.js
```diff
@@ -0,0 +1,23 @@
...
+  async runIfAuthenticated() {
+    // TO BE REMOVED. JUST CHECKING IF THE FLAG PARSING IS WORKING HERE :)
+    const parsed = this.parse(InitCommand);
+    logger.info('working here 🤙 flag:', parsed.flags);
+  }
+}
+
+InitCommand.description = 'Set up your development environment in your current folder.';
+
+InitCommand.flags = {
```
jeffladiray, line 15: You are missing the `help` option and description when running `forest init --help` :)

src/commands/init.js
```diff
@@ -0,0 +1,23 @@
...
+  project: flags.string({
+    char: 'p',
+    description: 'The name of the project you want to init.',
+    default: null,
```
jeffladiray, line 19: [Question] I'm not sure about this `default: null`. Might need to check, but I guess if you don't put this, you would simply not get the option if not given instead of getting `project=null`

src/commands/init.js
```diff
@@ -0,0 +1,23 @@
...
+    logger.info('working here 🤙 flag:', parsed.flags);
+  }
+}
+
+InitCommand.description = 'Set up your development environment in your current folder.';
+
+InitCommand.flags = {
+  project: flags.string({
+    char: 'p',
+    description: 'The name of the project you want to init.',
```
jeffladiray, line 18: See #42, but i think that we choosed to remove the final "." of the description

src/commands/init.js
```diff
@@ -0,0 +1,23 @@
...
+
+class InitCommand extends AbstractAuthenticatedCommand {
+  async runIfAuthenticated() {
+    // TO BE REMOVED. JUST CHECKING IF THE FLAG PARSING IS WORKING HERE :)
+    const parsed = this.parse(InitCommand);
+    logger.info('working here 🤙 flag:', parsed.flags);
+  }
+}
+
+InitCommand.description = 'Set up your development environment in your current folder.';
```
jeffladiray, line 13: See #42, but i think that we choosed to remove the final "." of the description

src/commands/init.js
adriguy: As discussed, to be consistent with other commands (and as it is not really useful as the help option is displayed only after having typed `--help`), let's not put it here. Do you agree?

src/commands/init.js
adriguy: As discussed, we keep it like this and we'll homogenise the other commands later on.

src/commands/init.js
adriguy: As discussed, we keep it like this and we'll homogenise the other commands later on.

src/commands/init.js
adriguy: Ok I remove it - let's see later on if we'll need it.

PR #54, opened 2020-04-30
fix: test configuration with tokens.

slimee:


src/abstract-authenticated-command.js
```diff
@@ -8,27 +8,24 @@ class AbstractAuthenticatedCommand extends Command {
     if (!authenticator.getAuthToken()) {
       logger.info('Login required.');
       await authenticator.tryLogin({});
+      if(!authenticator.getAuthToken()) this.displayLoginMessageAndQuit();
```
slimee, line 11: this is enough to stop when not logged. Adding a line here instead of wrapping the rest of the code.

src/services/authenticator.js
```diff
@@ -27,9 +27,8 @@ function Authenticator() {
     }
   };
 
-  this.pathToForestrc = `${os.homedir()}/.forestrc`;
-
-  this.saveToken = (token) => fs.writeFileSync(this.pathToForestrc, token);
+  this.saveToken = (token, path = process.env.TOKEN_PATH || os.homedir()) => fs
+    .writeFileSync(`${path}/.forestrc`, token);
```
slimee, line 31: Here I add the use of "TOKEN_PATH" variable. It allow tests that do not specify token mock (token:'any') to work with tokens in files. TOKEN_PATH is here to avoid overload our real token path.

test/commands/test-cli-auth-token.js
```diff
@@ -1,11 +1,30 @@
...
+    return token;
+  },
+  getTokenPath,
+  clearTokenPath,
   mockToken: (behavior) => {
     if (behavior !== null) {
       authenticator.getAuthTokenBack = authenticator.getAuthToken;
       authenticator.getAuthToken = () => behavior;
     }
+    clearTokenPath();
```
slimee, line 27: all this stuff is coming from authenticator.js test.
I've set it here because it will be reused by other tests.
It's the main job of this PR: merge token handling between "testCli" tests and "authenticator" tests.

test/fixtures/env.js
```diff
@@ -1,11 +1,11 @@
 module.exports = {
   testEnv: {
     FOREST_URL: 'http://localhost:3001',
-    TOKEN_PATH: '.',
+    TOKEN_PATH: './test/services/tokens',
   },
   testEnv2: {
     FOREST_URL: 'http://localhost:3001',
-    TOKEN_PATH: '.',
+    TOKEN_PATH: './test/services/tokens',
```
slimee, line 8: common path, we replace '.' by a path to not pollute the root of the project.
Note: './test/services/tokens' is in gitignore 👍

test/services/authenticator.test.js
```diff
@@ -1,84 +1,68 @@
...
-const fsExtra = require('fs-extra');
 const authenticator = require('../../src/services/authenticator');
-
-const key = 'test-token-key';
-const tokenPath = './test/services/tokens';
-const storeToken = (fileName, expiresIn = '14 days') => {
-  const token = jwt.sign({}, key, { expiresIn });
-  fs.writeFileSync(`${tokenPath}/${fileName}`, token);
-  return token;
-};
```
slimee, line 12: we just steal the code from here to set it into test-cli.
Like that, methods can de reused in test-cli-mock-token.

PR #55, opened 2020-05-04
feat(init-command): add the project selection and validation 2nd step (#4wmb1j)

adriguy:
https://app.clickup.com/t/4wmb1j


test/commands/branch.test.js
```diff
@@ -109,7 +109,6 @@ describe('branch', () => {
 
         describe('with a valid projectId', () => {
           it('should display a switch to new branch message', () => testCli({
-            print: true,
```
larcin, line 112: [Apologies] My bad on forgetting this one...

src/services/error-manager.js
```diff
@@ -0,0 +1,15 @@
...
+  if (!rawError.status) {
+    error = rawError;
+  } else {
+    error = ApiErrorDeserializer.deserialize(rawError);
+  }
+
+  return error.message ? error.message : 'Oops something went wrong';
+}
+
+module.exports = handleError;
```
larcin, line 15: [MISC] I have kind of an issue with this: this fie is called `error-manager` but directly exports a function.

Indeed it does not look like a service but more of a utility function file.

What about moving this file to `src/utils`, call it just `error.js` and expose the function as suggested:

```suggestion
module.exports = { handleError };
```

This way you can import it like this:

```
const { handleError } = require('../utils/error');
```

src/services/error-manager.js
```diff
@@ -0,0 +1,15 @@
+const ApiErrorDeserializer = require('../deserializers/api-error');
+
+function handleError(rawError) {
+  let error;
+  // NOTICE: We check if the errors are from the API or if they've been thrown internally.
+  if (!rawError.status) {
+    error = rawError;
+  } else {
+    error = ApiErrorDeserializer.deserialize(rawError);
```
larcin, line 9: [Error] I tried it without the backend PR and the 404 returned by the backend sends HTML (this is because of the un-existing route)

Thus we'll probably need to keep a try catch here to avoid badly formatted backend issues.

```
✖ Analyzing your setup
> SyntaxError: Unexpected token < in JSON at position 0
```

src/services/logger.js
```diff
@@ -1,6 +1,48 @@
...
-winston.remove(winston.transports.Console);
-winston.add(winston.transports.Console, { colorize: true });
+class Logger {
+  constructor(silent) {
+    this.silent = silent;
+    this.spinner = null;
+  }
 
-module.exports = winston;
+  pauseSpinner() {
```
larcin, line 9: [Spinner] I have some issues with the spinner that I'd like to discuss live:
- Where does it come from?
- Is this necessarily taking a promise like this?
- Pause/continue naming?

adriguy: As discussed together, let's take care of this loader refactoring once our init command is complete and ready to have the whole picture.

PR #59, opened 2020-05-12
Feature/dwo init command database setup

larcin:


src/commands/init.js
```diff
@@ -24,11 +40,38 @@ function handleInitError(rawError) {
   }
 }
 
+async function handleDatabaseConfiguration() {
+  if (process.env.DATABASE_URL) {
+    return null;
+  }
+
+  logger.success('✅ Checking your database setup');
```
adriguy, line 48: If we put your step in a spinner too, we won't need this log.
Let's talk about it to make all step success messages homogeneous.
image

src/commands/init.js
```diff
@@ -24,11 +40,38 @@ function handleInitError(rawError) {
   }
 }
 
+async function handleDatabaseConfiguration() {
+  if (process.env.DATABASE_URL) {
+    return null;
+  }
```
adriguy, line 46: I think we want the success message "✅Checking your database setup" even when we figured out that there was already a DATABASE_URL set up.

src/services/prompter/database-prompter.js
```diff
@@ -0,0 +1,209 @@
...
+    this.handleSchema();
+    this.handleHostname();
+    this.handlePort();
+    this.handleUser();
+    this.handlePassword();
+    this.handleSsl();
+    this.handleMongodbSrv();
+  }
+
+  // NOTICE: This is not used but required to stay iso with lumber for now
```
adriguy, line 33: 👏

src/commands/init.js
larcin: As said in private, this will be handle in next PR about spinner

PR #60, opened 2020-05-12
feat(init-command): create the user development environment (#4wmbcj)

adriguy:
https://app.clickup.com/t/4wmbcj

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -7,12 +7,15 @@ const spinners = require('../services/spinners');
 const logger = require('../services/logger');
 const ProjectManager = require('../services/project-manager');
 const DatabasePrompter = require('../services/prompter/database-prompter');
+const EnvironmentManager = require('../services/environment-manager');
 const { handleError } = require('../utils/error');
 const { buildDatabaseUrl } = require('../utils/database-url');
 
 const ERROR_MESSAGE_PROJECT_IN_V1 = 'This project does not support branches yet. Please migrate your environments from your Project settings first.';
 const ERROR_MESSAGE_NOT_ADMIN_USER = "You need the 'Admin' role to create a development environment on this project.";
 const ERROR_MESSAGE_PROJECT_NOT_FOUND = 'Your project was not found. Please check your environment secret.';
+const ERROR_MESSAGE_NO_PRODUCTION_OR_REMOTE_ENVIRONMENT = 'You cannot create your development environment until this project has either a remote or a production environment.';
+const ERROR_MESSAGE_ENVIRONMENT_OWNER_UNICITY = 'You already have a development environment on this project.';
 
 const OPTIONS_DATABASE = [
   'dbDialect',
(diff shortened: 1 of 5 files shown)


src/commands/init.js
```diff
@@ -1,14 +1,18 @@
 const { flags } = require('@oclif/command');
+const inquirer = require('inquirer');
 const AbstractAuthenticatedCommand = require('../abstract-authenticated-command');
 const withCurrentProject = require('../services/with-current-project');
 const spinners = require('../services/spinners');
 const logger = require('../services/logger');
 const ProjectManager = require('../services/project-manager');
 const { handleError } = require('../utils/error');
+const EnvironmentManager = require('../services/environment-manager');
```
larcin, line 9: [Code style] It's a good practice to keep all services import together

src/commands/init.js
```diff
@@ -19,33 +23,85 @@ function handleInitError(rawError) {
...
   }
 }
 
+function validateEndpoint(input) {
+  if (!/^https?:\/\/.*/i.test(input)) {
+    return 'Application input must be a valid url.';
+  }
+  if (!/^http((s:\/\/.*)|(s?:\/\/(localhost|127\.0\.0\.1).*))/i.test(input)) {
+    return 'HTTPS protocol is mandatory, except for localhost and 127.0.0.1.';
+  }
```
larcin, line 43: [Code Style] Those two regular expressions could be stored into constants at the start of the file

src/commands/init.js
```diff
@@ -1,14 +1,21 @@
...
 const { handleError } = require('../utils/error');
 
 const ERROR_MESSAGE_PROJECT_IN_V1 = 'This project does not support branches yet. Please migrate your environments from your Project settings first.';
 const ERROR_MESSAGE_NOT_ADMIN_USER = "You need the 'Admin' role to create a development environment on this project.";
 const ERROR_MESSAGE_PROJECT_NOT_FOUND = 'Your project was not found. Please check your environment secret.';
+const ERROR_MESSAGE_NO_PRODUCTION_OR_REMOTE_ENVIRONMENT = 'You cannot create your development environment until this project has either a remote or a production environment.';
+const ERROR_MESSAGE_ENVIRONMENT_OWNER_UNICITY = 'You already have a development environment on this project.';
+
+const URL_VALIDATION_REGEX = /^https?:\/\/.*/i;
+const HTTPS_VALIDATION_REGEX = /^http((s:\/\/.*)|(s?:\/\/(localhost|127\.0\.0\.1).*))/i;
```
larcin, line 18: [Naming] Being a bit picky here, but when writing constants we tend to put "what it is first" (that seems to be a convention here, even if that doesn't sound english):
```suggestion
const VALIDATION_REGEX_URL = /^https?:\/\/.*/i;
const VALIDATION_REGEX_HTTPS = /^http((s:\/\/.*)|(s?:\/\/(localhost|127\.0\.0\.1).*))/i;
```

adriguy: As you wish!

PR #61, opened 2020-05-12
feat(branch): add switch command

rap2hpoutre:
See: https://app.clickup.com/t/56nznf


src/commands/switch.js
```diff
@@ -0,0 +1,99 @@
...
+        ].map((currentBranch) => currentBranch.name),
+      }]);
+
+      return branch;
+    } catch (error) {
+      const customError = BranchManager.handleBranchError(error);
+
+      return this.error(customError);
+    }
+  }
```
arnaud-moncel, line 29: I think this function can be put in an helper like `with-current-project`. it should be reused by the deploy command IMO.

src/commands/switch.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    return config;
+  }
+
+  async runIfAuthenticated() {
+    try {
+      const envSecret = process.env.FOREST_ENV_SECRET;
+      const config = await this.getConfig(envSecret);
+      const branches = await BranchManager.getBranches(envSecret) || [];
```
arnaud-moncel, line 62: Here the command crash when i use it outside a project.
If you get the envSecret from the config it should be work.
```suggestion
      const branches = await BranchManager.getBranches(config.envSecret) || [];
```

src/commands/switch.js
```diff
@@ -0,0 +1,99 @@
...
+        .getDevelopmentEnvironmentForUser(config.projectId);
+      config.envSecret = environment.secretKey;
+    }
+
+    return config;
+  }
+
+  async runIfAuthenticated() {
+    try {
+      const envSecret = process.env.FOREST_ENV_SECRET;
```
arnaud-moncel, line 60: This line can be moved inside the `getConfig` function IMO

src/services/branch-manager.js
```diff
@@ -42,6 +43,16 @@ function createBranch(branchName, environmentSecret) {
...
 }
 
+function switchBranch({ id }, environmentSecret) {
+  const authToken = authenticator.getAuthToken();
+
+  return agent
+    .put(`${serverHost()}/api/environments`)
+    .set('Authorization', `Bearer ${authToken}`)
+    .set('forest-secret-key', `${environmentSecret}`)
+    .send(EnvironmentSerializer.serialize({ currentBranchId: id }));
```
arnaud-moncel, line 53: It's pretty strange to call the serializer here to build an environment just with a `currentBranchId` no ?

src/services/branch-manager.js
```diff
@@ -42,6 +43,16 @@ function createBranch(branchName, environmentSecret) {
     .send({ branchName: encodeURIComponent(branchName) });
 }
 
+function switchBranch({ id }, environmentSecret) {
+  const authToken = authenticator.getAuthToken();
+
+  return agent
+    .put(`${serverHost()}/api/environments`)
```
arnaud-moncel, line 50: Call on environment route inside `BranchManager` is pretty strange no ?

rap2hpoutre: We could consider it valid, however I'm OK to change my mind! 

The actual action is related to branch: the action is indeed **`switch branch`**, so it could be in the **branch service**. The `switchBranch` function have to **patch** an environment to update its current branch.

When I look at environment service, I see:

- /api/projects/xxx//environments
- /api/environments/
- /api/deployment-requests/

So it's not consistent either with the service name. I don't know if it's an issue.

Still, I do not have a real preference: I guess it could be in environment service too. So, if you have a preference, I would follow yours! It's up to you

src/services/branch-manager.js
rap2hpoutre: It has been added to be consistent with other API call: AFAIK, each API call have to use a serializer, example here:

```
  this.updateEnvironment = async () => {
    const authToken = authenticator.getAuthToken();
    return agent
      .put(`${serverHost()}/api/environments/${config.environmentId}`)
      .set('Authorization', `Bearer ${authToken}`)
      .send(EnvironmentSerializer.serialize({
        name: config.name,
        apiEndpoint: config.url,
      }));
  };
```

rap2hpoutre: Side note: `PUT` is currently used, a REST keeper would consider it should have been a `PATCH` since it already accepts partial update, but I did not want to refactor this whole part.

src/commands/switch.js
rap2hpoutre: Thanks for this suggestion. At first glance, I thought I agree but since then I changed my mind. This method is currently specific to this case.

```js
message: 'Select the branch you want to set current',
```

```js
// NOTICE: Current branch should be last displayed branch.
```

We could consider avoiding a premature blind factorization: we will factorize (if required) when we need it, it would be better crafted. In other word, the developer that code the deploy command should not copy/paste if it's the exact same code: he should consider factorize it.

arnaud-moncel: Ok keep it like that we will see.

src/services/branch-manager.js
arnaud-moncel: As we discuss it's a little bit overkill but keep it like that.

PR #66, opened 2020-05-14
chore: spinner refactoring first draft

larcin:
Improve the way we use the spinner, it's now a singleton instance that you can use either synchronously or asynchronously

Only drawback to this implem: there can only be one spinner at a time


src/services/spinner.js
```diff
@@ -0,0 +1,85 @@
...
+  runOnPromise(options, promise) {
+    this.start(options);
+
+    return promise
+      .then((result) => {
+        this.success(options);
+        return result;
+      })
+      .catch((error) => {
+        this.fail({ text: error });
```
adriguy, line 80: Here I guess you should throw the error so that we'll catch it in the job code ("code métier" ^^)
```suggestion
        this.fail({ text: error });
        throw error;
```

src/services/spinner.js
```diff
@@ -0,0 +1,99 @@
...
+  stop() {
+    if (!this.isRunning()) {
+      throw Error('No spinner is running.');
+    }
+
+    this.spinnies.stopAll();
+    this.currentSpinnerOptions = null;
+  }
+
+  success() {
```
adriguy, line 37: Judging from the doc, I think it's possible to add options here too (to update the text for instance).
```suggestion
  success(options = this.currentSpinnerOptions) {
```

src/services/with-current-project.js
```diff
@@ -21,14 +20,13 @@ module.exports = async function withCurrentProject(config) {
...
 
-    logger.pauseSpinner();
     const response = await inquirer.prompt([{
       name: 'project',
       message: 'Select your project',
       type: 'list',
       choices: projects.map((project) => ({ name: project.name, value: project.id })),
     }]);
-    logger.continueSpinner();
+
```
adriguy, line 29: Newline mistake?

test/services/spinner.test.js
```diff
@@ -0,0 +1,51 @@
...
+        } catch (e) {
+          message = e.message;
+        }
+
+        expect(message).toBe('wrong');
+        expect(spy).toHaveBeenCalledWith({ text: error });
+      });
+    });
+  });
+});
```
adriguy, line 51: Awesome tests!
Do you plan creating some on pause / continue? 
Just asking here, it might be not really relevant.

PR #69, opened 2020-05-16
feat(init-command): auto-fill or display environment variables (#4wmbgy)

adriguy:
https://app.clickup.com/t/4wmbgy

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "bluebird": "3.5.2",
     "chalk": "4.0.0",
     "cli-table": "github:Automattic/cli-table#master",
+    "clipboardy": "2.3.0",
     "dotenv": "8.2.0",
     "inquirer": "6.2.0",
     "joi": "14.3.1",
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -140,7 +140,10 @@ function commentExistingVariablesInAFile(fileData, config) {
       'DATABASE_SSL=': '# DATABASE_SSL=',
     };
   }
-  const variablesToCommentRegex = new RegExp(Object.keys(variablesToComment).join('|'), 'g');
+  const variablesToCommentRegex = new RegExp(
+    Object.keys(variablesToComment).map((key) => `((?<!# )${key})`).join('|'),
+    'g',
+  );
   return fileData.replace(variablesToCommentRegex, (match) => variablesToComment[match]);
 }
 
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -45,7 +45,10 @@ const OPTIONS_DATABASE = [
 
 const VALIDATION_REGEX_URL = /^https?:\/\/.*/i;
 const VALIDATION_REGEX_HTTPS = /^http((s:\/\/.*)|(s?:\/\/(localhost|127\.0\.0\.1).*))/i;
+const SPLIT_URL_REGEX = new RegExp('(\\w+)://([\\w\\-\\.]+)(:(\\d+))?');
 
+const PROMPT_MESSAGE_AUTO_FILLING_ENV_FILE = 'Do you want your current folder `.env` file to be completed automatically with your environment variables?';
+const PROMPT_MESSAGE_AUTO_CREATING_ENV_FILE = 'Do you want a new `.env` file (containing your environment variables) to be automatically created in your current folder?';
 const ENV_VARIABLES_AUTO_FILLING_PREFIX = '\n\n# ℹ️ The content below was automatically added by the `forest init` command ⤵️\n';
 
 function handleInitError(rawError) {
@@ -94,14 +97,11 @@ function validateEndpoint(input) {
 }
 
 function getApplicationPortFromCompleteEndpoint(endpoint) {
-  if (endpoint.startsWith('http://localhost:') || endpoint.startsWith('http://127.0.0.1:')) {
-    return endpoint.substring(17);
-  }
-  return null;
+  return endpoint.match(SPLIT_URL_REGEX)[4];
 }
 
-async function getContentToAddInDotenvFile(config) {
-  const authSecret = await generateKey();
+function getContentToAddInDotenvFile(config) {
+  const authSecret = generateKey();
   let contentToAddInDotenvFile = '';
 
   if (config.applicationPort) {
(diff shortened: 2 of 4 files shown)


src/commands/init.js
```diff
@@ -80,6 +83,96 @@ async function handleDatabaseConfiguration() {
...
+  }
+  if (config.databaseSchema) {
+    contentToAddInDotenvFile += `DATABASE_SCHEMA=${config.databaseSchema}\n`;
+  }
+  if (config.databaseSSL !== undefined) {
+    contentToAddInDotenvFile += `DATABASE_SSL=${config.databaseSSL}\n`;
+  }
+  contentToAddInDotenvFile += `FOREST_AUTH_SECRET=${authSecret}\n`;
+  contentToAddInDotenvFile += `FOREST_ENV_SECRET=${config.forestEnvSecret}`;
+  return contentToAddInDotenvFile;
```
larcin, line 121: [Performance] Using `+=` to build string is a really bad practice in many coding languages (as it has to reallocate memory space for each addition) thus using `concat` should be preferred. However I'm not entirely sure that this applies to js

larcin: After a quick search, it seems indeed that that performance issue doesn't happen in js. Let's stick with the `+=` then.

src/commands/init.js
```diff
@@ -80,6 +83,96 @@ async function handleDatabaseConfiguration() {
...
+function commentExistingVariablesInAFile(fileData, config) {
+  let variablesToComment = {
+    'FOREST_AUTH_SECRET=': '# FOREST_AUTH_SECRET=',
+    'FOREST_ENV_SECRET=': '# FOREST_ENV_SECRET=',
+  };
+  if (config.applicationPort) {
+    variablesToComment = {
+      ...variablesToComment,
+      'APPLICATION_PORT=': '# APPLICATION_PORT=',
+    };
```
larcin, line 133:
```suggestion
    variablesToComment['APPLICATION_PORT='] = '# APPLICATION_PORT=';
```

src/commands/init.js
```diff
@@ -80,6 +83,96 @@ async function handleDatabaseConfiguration() {
...
+  }
+  if (config.databaseUrl) {
+    variablesToComment = {
+      ...variablesToComment,
+      'DATABASE_URL=': '# DATABASE_URL=',
+      'DATABASE_SCHEMA=': '# DATABASE_SCHEMA=',
+      'DATABASE_SSL=': '# DATABASE_SSL=',
+    };
```
larcin, line 141:
```suggestion
    variablesToComment['DATABASE_URL='] = '# DATABASE_URL=';
    variablesToComment['DATABASE_SCHEMA='] = '# DATABASE_SCHEMA=';
    variablesToComment['DATABASE_SSL='] = '# DATABASE_SSL=';
```

src/commands/init.js
```diff
@@ -80,6 +83,96 @@ async function handleDatabaseConfiguration() {
...
+  }
+  return true;
+}
+
+function getApplicationPortFromCompleteEndpoint(endpoint) {
+  if (endpoint.startsWith('http://localhost:') || endpoint.startsWith('http://127.0.0.1:')) {
+    return endpoint.substring(17);
+  }
+  return null;
+}
```
larcin, line 101: [Misc] Does that mean that we choose not to handle any other local endpoints?

package.json
```diff
@@ -23,6 +23,7 @@
     "bluebird": "3.5.2",
     "chalk": "4.0.0",
     "cli-table": "github:Automattic/cli-table#master",
+    "clipboardy": "2.3.0",
```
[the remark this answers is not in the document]
adriguy: After searching on forums for the best way to copy into the clipboard (not from a browser) and that is working for several OS, I did not found an easy solution other than using a dependency.
And to be truly honest, I did not find a lot of them. I picked this one, checked if it worked well, et voilà!
So if you have in mind other ways to do it, or think it deserves more research on what is the best clipboard dependency for us, I am totally open to do so 🙂

src/utils/key-generator.js
```diff
@@ -0,0 +1,10 @@
+const P = require('bluebird');
+const crypto = require('crypto');
+
+const randomBytes = P.promisify(crypto.randomBytes);
```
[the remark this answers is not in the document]
adriguy: I copied it from lumber for the main part.
Would you like me to try to get rid of it?

src/commands/init.js
adriguy: I actually hesitate between `+=` and `concat`.
I choose `+=` because it looked cleaner to me.
Thanks for the research on the subject!

larcin:
There seems to be an issue, if I'm launching the init from another folder, the .env gets created anyway. Shouldn't we either (or both) check that the current folder is a relevant one (through presence of a `package.json` file for example) or ask the user if he wants to have its `.env` created?

src/commands/init.js
```diff
@@ -80,6 +83,99 @@ async function handleDatabaseConfiguration() {
...
+
+async function getContentToAddInDotenvFile(config) {
+  const authSecret = await generateKey();
+  let contentToAddInDotenvFile = '';
+
+  if (config.applicationPort) {
+    contentToAddInDotenvFile += `APPLICATION_PORT=${config.applicationPort}\n`;
+  }
```
larcin, line 109: [Misc] We don't handle the `APPLICATION_URL`?

package.json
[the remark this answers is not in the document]
larcin: Oh no, this one sounds perfect, I was just curious about the story behind it (whether it was from an old usage in previous experience, inherited from Lumber, or just the fruit of your research)

src/commands/init.js
adriguy: 👍 
I was not sure what was the best way to do it between those 2.
Do you prefer this one because it's more compact?

src/commands/init.js
adriguy: What other local endpoint could exist?
Maybe `::1`? 🤔

src/commands/init.js
adriguy: Well, I am not sure it is needed, as it not specified in the lumber generated projects.
Do you confirm?

adriguy:
> There seems to be an issue, if I'm launching the init from another folder, the .env gets created anyway. Shouldn't we either (or both) check that the current folder is a relevant one (through presence of a `package.json` file for example) or ask the user if he wants to have its `.env` created?

Well, as the description of the command is:
image
I considered that was not an issue.
But now that you bring it up, I think you're right, we should deal with this case.

Another similar case could be a problem: if a user is running this command (with the `--project` option for instance) but in another project/environment repo.
Do you think we should ask for a confirmation even when we fill the `.env` file? If the answer is no, we just display the env variables.

src/commands/init.js
adriguy: I'll correct it to match all endpoint ports.

src/utils/key-generator.js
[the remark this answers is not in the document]
adriguy: I'll do it 👍

PR #75, opened 2020-05-20
fix(spinner): avoid promise failures on paused spinner

larcin:
## Pull Request checklist:

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [ ] Create automatic tests
- [ ] No automatic tests failures
- [ ] Review my own code (indentation, syntax, style, simplicity, readability)
- [ ] Wonder if you can improve the existing code

### DIFF
--- a/src/services/spinner.js
+++ b/src/services/spinner.js
@@ -108,7 +108,10 @@ class Spinner {
         return result;
       })
       .catch((error) => {
-        this.fail({ text: error });
+        // NOTICE: Only trigger the fail if the spinner is running (ie. not paused)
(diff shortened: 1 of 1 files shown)


larcin:
> Good to me!
> Should you do the same in the `then` just above to prevent similar issue when trying to do a `this.success()` whereas the spinner is paused?

I don't think so. This only allows us to handle errors thanks happen outside of the flow. Success are only intended to be used through a correct flow.

PR #81, opened 2020-05-20
fix(init-command): fix error handling

adriguy:
## Pull Request checklist:

- [ ] Create automatic tests
- [x] No automatic tests failures

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -61,9 +61,7 @@ class InitCommand extends AbstractAuthenticatedCommand {
 
   async projectSelection() {
     const parsed = this.parse(InitCommand);
-    spinner.pause();
     this.config = await withCurrentProject({ ...envConfig, ...parsed.flags });
-    spinner.continue();
   }
 
   async projectValidation() {
--- a/src/services/with-current-project.js
+++ b/src/services/with-current-project.js
@@ -24,6 +24,7 @@ module.exports = async function withCurrentProject(config) {
       return { ...config, projectId: projects[0].id };
     }
 
+    // NOTICE: If a spinner is running, it has to be paused during the prompt and resumed after.
     const existingSpinner = spinner.isRunning();
     if (existingSpinner) spinner.pause();
     const response = await inquirer.prompt([{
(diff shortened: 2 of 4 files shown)


src/commands/init.js
```diff
@@ -61,9 +61,7 @@ class InitCommand extends AbstractAuthenticatedCommand {
 
   async projectSelection() {
     const parsed = this.parse(InitCommand);
-    spinner.pause();
```
adriguy, line 64: I remove it from here to be able to have the spinner under error:
image
This is what I had if I leave it here. 
And moving it to the `with-current-project`:
image

src/utils/error.js
```diff
@@ -6,6 +6,9 @@ function handleError(rawError) {
   let error;
   // NOTICE: We check if the errors are from the API (have a status) or if thrown internally.
   if (rawError.status) {
+    if (rawError.status === 500) {
+      return UNEXPECTED_ERROR_MESSAGE;
+    }
```
adriguy, line 11: I took the opportunity to be sure we'll never have this kind of message:
image

adriguy:
@larcin there is one thing that I noticed that is not perfect:
image
The project not found is displayed under the "analyzing your setup"...
But this is because of the way the `with-current-project` is build, if a --project is given it return the id without checking if the project exists.
 I am not sure we want to change this unless you have an easy solution 🤔

src/services/with-current-project.js
```diff
@@ -20,12 +24,15 @@ module.exports = async function withCurrentProject(config) {
       return { ...config, projectId: projects[0].id };
     }
 
+    const existingSpinner = spinner.isRunning();
+    if (existingSpinner) spinner.pause();
```
larcin, line 28: [MISC] You could add a comment here to specify that this is intended for usage outside of a spinner context

src/utils/error.js
larcin: [MISC] I'm not quite sure that we should do something like that. 500 errors shouldn't exist, that mens we should look for it directly when needed.

IMO we should have those displayed

PR #87, opened 2020-05-28
fix: homogenize all displayed error messages style

adriguy:
https://app.clickup.com/t/4ygzep

## Pull Request checklist:

- [ ] Create automatic tests
- [x] No automatic tests failures

### DIFF
--- a/src/commands/branch.js
+++ b/src/commands/branch.js
@@ -4,6 +4,7 @@ const AbstractAuthenticatedCommand = require('../abstract-authenticated-command'
 const BranchManager = require('../services/branch-manager');
 const ProjectManager = require('../services/project-manager');
 const withCurrentProject = require('../services/with-current-project');
+const logger = require('../services/logger');
 const envConfig = require('../config');
 
 class BranchCommand extends AbstractAuthenticatedCommand {
@@ -19,7 +20,8 @@ class BranchCommand extends AbstractAuthenticatedCommand {
     } catch (error) {
       const customError = BranchManager.handleBranchError(error);
 
-      return this.error(customError);
+      logger.error(customError);
+      return this.exit(2);
     }
     return null;
   }
@@ -32,7 +34,8 @@ class BranchCommand extends AbstractAuthenticatedCommand {
     } catch (error) {
       const customError = BranchManager.handleBranchError(error);
 
-      return this.error(customError);
+      logger.error(customError);
+      return this.exit(2);
     }
   }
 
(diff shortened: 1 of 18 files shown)


src/commands/branch.js
```diff
@@ -19,7 +20,8 @@ class BranchCommand extends AbstractAuthenticatedCommand {
     } catch (error) {
       const customError = BranchManager.handleBranchError(error);
 
-      return this.error(customError);
+      logger.error(customError);
+      return this.exit(2);
```
larcin, line 24: [Misc] Why did you make an `exit(2)` here?

larcin: Same question for all `exit(2)` across this file

src/commands/environments/copy-layout.js
```diff
@@ -1,14 +1,15 @@
...
 const AbstractAuthenticatedCommand = require('../../abstract-authenticated-command');
+const EnvironmentManager = require('../../services/environment-manager');
 const withCurrentProject = require('../../services/with-current-project');
+const logger = require('../../services/logger');
 const envConfig = require('../../config');
 
 class CopyLayoutCommand extends AbstractAuthenticatedCommand {
   async runIfAuthenticated() {
-    const logError = this.error.bind(this);
+    const oclifExit = this.exit.bind(this);
```
larcin, line 12: 👏

src/commands/environments/create.js
```diff
@@ -19,11 +20,8 @@ class CreateCommand extends AbstractAuthenticatedCommand {
         const errorData = JSON.parse(error.response.text);
         if (errorData && errorData.errors && errorData.errors.length
           && errorData.errors[0] && errorData.errors[0].detail) {
-          this.error(errorData.errors[0].detail, {
-            code: errorData.errors[0].status,
-            exit: 1,
-          });
-          return;
+          logger.error(errorData.errors[0].detail);
+          this.exit(1);
```
larcin, line 24: [MISC] If you want to stay iso with what existed (and avoid the following `throw error`, you should return here.

src/services/job-state-checker.js
```diff
@@ -5,10 +5,11 @@ const { promisify } = require('util');
 const jobDeserializer = require('../deserializers/job');
 const config = require('../config');
 const { getAuthToken } = require('./authenticator');
+const logger = require('../services/logger');
```
larcin, line 8:
```suggestion
const logger = require('./logger');
```

src/commands/branch.js
adriguy: As discussed, it is the default exit code thrown by `this.error`...
So to stay iso and satisfy the current tests, I had to specify the exit 2.

src/commands/environments/create.js
adriguy: As discussed, after the `this.exit(1)`, all code added is flagged as unreachable by my IDE.
So let's keep it that way!

PR #99, opened 2020-07-29
fix(apply:shema)!: require authentication

ghusse:
Linked to CU-6jpyjv

Command schema:apply now requires authentication, to avoid 401 errors.

## Pull Request checklist:

- [x] Create automatic tests
- [ ] No automatic tests failures

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ test/services/tokens
 # IDE
 /*.iml
 .idea
+/.vscode
--- a/src/commands/schema/apply.js
+++ b/src/commands/schema/apply.js
@@ -1,21 +1,29 @@
-const { Command, flags } = require('@oclif/command');
+const { flags } = require('@oclif/command');
 const fs = require('fs');
 const path = require('path');
 const Joi = require('joi');
 const SchemaSerializer = require('../../serializers/schema');
 const SchemaSender = require('../../services/schema-sender');
 const JobStateChecker = require('../../services/job-state-checker');
+const AbstractAuthenticatedCommand = require('../../abstract-authenticated-command');
 const logger = require('../../services/logger');
+const authenticator = require('../../services/authenticator');
 
-class ApplyCommand extends Command {
-  async run() {
+class ApplyCommand extends AbstractAuthenticatedCommand {
+  async runIfAuthenticated() {
     const oclifExit = this.exit.bind(this);
     const { flags: parsedFlags } = this.parse(ApplyCommand);
     const serializedSchema = this.readSchema();
     const secret = this.getEnvironmentSecret(parsedFlags);
+    const authToken = authenticator.getAuthToken();
 
     this.log('Sending "./.forestadmin-schema.json"...');
-    const jobId = await new SchemaSender(serializedSchema, secret, oclifExit).perform();
+    const jobId = await new SchemaSender(
+      serializedSchema,
+      secret,
+      oclifExit,
+      authToken,
+    ).perform();
 
     if (jobId) {
       await new JobStateChecker('Processing schema', oclifExit).check(jobId);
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ test/services/tokens
 # IDE
 /*.iml
 .idea
+/.vscode
```
arnaudbesnier, line 14: This is a kind of Pull Request pollution as it does not serves the purpose of the current intent.
I know it is nothing (1 line of code) but it should not be here in my opinion, and should have it's dedicated Pull Request with its own commit message that describes a new intent.

src/services/schema-sender.js
```diff
@@ -3,11 +3,23 @@ const agent = require('superagent-promise')(require('superagent'), P);
...
 
-function SchemaSender(serializedSchema, secret, oclifExit) {
+/**
+ * @class
+ * @param {string} serializedSchema
+ * @param {string} secret
+ * @param {(code: number) => void} oclifExit
+ * @param {string} authToken
+ */
+function SchemaSender(serializedSchema, secret, oclifExit, authToken) {
```
arnaudbesnier, line 13:
```suggestion
function SchemaSender(serializedSchema, secret, authToken, oclifExit) {
```
would probably be a better signature as `secret` and `authToken` have a similar authentication/identification purpose and `oclifExit` seems here to handle errors.

src/commands/schema/apply.js
```diff
@@ -1,21 +1,29 @@
...
 
-class ApplyCommand extends Command {
-  async run() {
+class ApplyCommand extends AbstractAuthenticatedCommand {
+  async runIfAuthenticated() {
     const oclifExit = this.exit.bind(this);
     const { flags: parsedFlags } = this.parse(ApplyCommand);
     const serializedSchema = this.readSchema();
     const secret = this.getEnvironmentSecret(parsedFlags);
+    const authToken = authenticator.getAuthToken();
```
arnaudbesnier, line 18: I know the function is named `getAuthToken`, but I personally prefer to use complete words for variables names.
```suggestion
    const authenticationToken = authenticator.getAuthToken();
```
seems really more explicit to me (especially for juniors members that would have to dig into this code).
(Same comment for the `SchemaSender` constructor new argument name)

test/commands/schema/apply.test.js
```diff
@@ -2,134 +2,165 @@ const testCli = require('./../test-cli');
...
+              { out: 'Sending "./.forestadmin-schema.json"...' },
+              { out: 'The schema is the same as before, nothing changed.' },
+            ],
+          }));
+        });
+      });
     });
   });
+
+  describe('when the user is not logged in', () => {
```
arnaudbesnier, line 140: I would move this test on top of the file (before "when the user is logged in"),
because it is the first scenario to test in my opinion, not one of the last.

test/commands/schema/apply.test.js
```diff
@@ -2,134 +2,165 @@ const testCli = require('./../test-cli');
...
+    it('should as for the login/password and then send the schema', () => testCli({
```
arnaudbesnier, line 141:
```suggestion
    it('should ask for the login/password and then send the schema', () => testCli({
```
?

.gitignore
ghusse: Fixed by splitting this PR into 2 : #100 and #99

src/services/schema-sender.js
ghusse: Fixed in #100

src/commands/schema/apply.js
ghusse: Fixed in #100

test/commands/schema/apply.test.js
ghusse: Fixed in #100

test/commands/schema/apply.test.js
ghusse: Fixed in #100

PR #115, opened 2020-10-06
test: fix random test errors

arnaudbesnier:
## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "husky": "4.2.4",
     "jest": "24.9.0",
     "mock-stdin": "1.0.0",
-    "nock": "12.0.3",
+    "nock": "13.0.4",
     "semantic-release": "17.0.7",
     "semantic-release-slack-bot": "1.6.2",
     "simple-git": "1.65.0",
--- a/test/commands/schema/apply.test.js
+++ b/test/commands/schema/apply.test.js
@@ -155,19 +155,19 @@ describe('schema:apply', () => {
       });
     });
   });
-});
 
-describe('with forest server returning nothing', () => {
-  it('should exit with exit code 6', () => testCli({
-    file: {
-      name: '.forestadmin-schema.json',
-      content: forestadminSchema,
-    },
-    env: testEnv2,
-    token: 'any',
-    api: [postSchema500()],
-    command: () => ApplySchemaCommand.run([]),
-    std: [{ err: 'An error occured with the schema sent to Forest. Please contact support@forestadmin.com for further investigations.' }],
-    exitCode: 6,
-  }));
+  describe('with forest server returning nothing', () => {
+    it('should exit with exit code 6', () => testCli({
+      file: {
+        name: '.forestadmin-schema.json',
+        content: forestadminSchema,
+      },
+      env: testEnv2,
+      token: 'any',
+      api: [postSchema500()],
+      command: () => ApplySchemaCommand.run([]),
+      std: [{ err: 'An error occured with the schema sent to Forest. Please contact support@forestadmin.com for further investigations.' }],
+      exitCode: 6,
+    }));
+  });
 });
--- a/test/commands/test-cli-std.js
+++ b/test/commands/test-cli-std.js
@@ -14,7 +14,8 @@ module.exports = {
   },
   planifyInputs: (inputs, stdin) => {
     for (let i = 0; i < inputs.length; i += 1) {
-      setTimeout(() => stdin.send(`${inputs[i]}\n`), 500 + i * 100);
+      // Smelly code...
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "husky": "4.2.4",
     "jest": "24.9.0",
     "mock-stdin": "1.0.0",
-    "nock": "12.0.3",
+    "nock": "13.0.4",
```
arnaudbesnier, line 59: I updated nock during the investigations, it works well with this version, we should keep it.
I know this is something that should ideally be isolated in a dedicated contribution.

test/commands/schema/apply.test.js
```diff
@@ -155,19 +155,19 @@ describe('schema:apply', () => {
...
-      content: forestadminSchema,
-    },
-    env: testEnv2,
-    token: 'any',
-    api: [postSchema500()],
-    command: () => ApplySchemaCommand.run([]),
-    std: [{ err: 'An error occured with the schema sent to Forest. Please contact support@forestadmin.com for further investigations.' }],
-    exitCode: 6,
-  }));
+  describe('with forest server returning nothing', () => {
```
arnaudbesnier, line 159: [Improvement] A simple indentation of the existing test to move it in the main `describe`:
`describe('schema:apply', () => {`

yarn.lock
```diff
@@ -6274,14 +6279,14 @@ nice-try@^1.0.4:
   resolved "https://registry.yarnpkg.com/nice-try/-/nice-try-1.0.5.tgz#a3378a7696ce7d223e88fc9b764bd7ef1089e366"
   integrity sha512-1nh45deeb5olNY7eX82BkPO7SSxR5SSYJiPTrTdFUVYwAl8CKMA5N9PjTYkHiRjisVcxcQ1HXdLhx2qxxJzLNQ==
 
-nock@12.0.3:
-  version "12.0.3"
-  resolved "https://registry.yarnpkg.com/nock/-/nock-12.0.3.tgz#83f25076dbc4c9aa82b5cdf54c9604c7a778d1c9"
-  integrity sha512-QNb/j8kbFnKCiyqi9C5DD0jH/FubFGj5rt9NQFONXwQm3IPB0CULECg/eS3AU1KgZb/6SwUa4/DTRKhVxkGABw==
+nock@13.0.4:
```
arnaudbesnier, line 6282: `nock` upgrade consequences.

test/services/spinner.test.js
```diff
@@ -143,7 +145,6 @@ describe('spinner', () => {
         expect(spinner.currentSpinnerOptions).toBeNull();
       });
 
-
```
arnaudbesnier, line 146: [Code Style] Consistent test separations.
Nothing related to the tests fix.

test/commands/test-cli-std.js
```diff
@@ -14,7 +17,8 @@ module.exports = {
   },
   planifyInputs: (inputs, stdin) => {
     for (let i = 0; i < inputs.length; i += 1) {
-      setTimeout(() => stdin.send(`${inputs[i]}\n`), 500 + i * 100);
+      // WARNING: Smelly code..., could break if the test pre-loading time increases.
+      setTimeout(() => stdin.send(`${inputs[i]}\n`), TIME_TO_LOAD_TEST + i * TIME_TO_REACH_NEXT_PROMPT);
```
arnaudbesnier, line 21: Here is the dirty part... and the reason of the issues.
I think the inputs were send before the prompt requested them in some tests, so some random tests were waiting and timed out, or some nock call where never executed by the commands.

I increase the initial time to load from 500ms to 1000ms.
Open to better (quick) solutions.

test/services/spinner.test.js
```diff
@@ -94,6 +94,7 @@ describe('spinner', () => {
           spinner.continue();
           expect(spinner.isRunning()).toBe(true);
           expect(spinner.pausedSpinnerOptions).toBeNull();
+          spinner.pause();
```
arnaudbesnier, line 97: Prevent "open handles" detection.
See https://jestjs.io/docs/en/cli#--detectopenhandles from more information.

test/commands/test-cli-std.js
jeffladiray: Ok, so if I get it right, you increased the timeout for the next (following) input ?
Seems to solve the issue at least for now (I launched the test 25 times from using a bash script just to be sure).

This is indeed not a definitive fix, but the original implementation already seems to go that way, and with the current test code structure I can't really think of a better solution

jeffladiray (approved):
Launched using `for i in {1..25}; do yarn test; done` to detect potential flakiness, but it seems good for that part.

As stated on my comment, I can't find any better solution to fix this issue than what you did, or a rework of the test structure, so SGTM for now 👍

PR #122, opened 2020-11-20
fix(deploy-command): allow project to be passed in as argument using --projectId or -p

SteveBunlon:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -114,7 +114,7 @@ DeployCommand.flags = {
   }),
   projectId: flags.string({
     char: 'p',
-    description: 'Forest project ID.',
+    description: 'The id of the project you want to deploy.',
     default: null,
   }),
 };
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -112,6 +112,11 @@ DeployCommand.flags = {
     char: 'f',
     description: 'Skip deploy confirmation.',
   }),
+  projectId: flags.string({
+    char: 'p',
+    description: 'Forest project ID.',
+    default: null,
+  }),
 };
 
 DeployCommand.args = [{
--- a/test/commands/deploy.test.js
+++ b/test/commands/deploy.test.js
@@ -63,7 +63,7 @@ describe('deploy', () => {
       it('should not display the list of projects', () => testCli({
         env: testEnv2,
         token: 'any',
-        command: () => DeployCommand.run(['-p', projectId]),
+        command: () => DeployCommand.run(['-p', '82']),
         api: [
           getEnvironmentListValid(projectId),
           deployValid(),
--- a/test/commands/deploy.test.js
+++ b/test/commands/deploy.test.js
@@ -63,9 +63,8 @@ describe('deploy', () => {
       it('should not display the list of projects', () => testCli({
         env: testEnv2,
         token: 'any',
-        command: () => DeployCommand.run([]),
+        command: () => DeployCommand.run(['-p', projectId]),
         api: [
-          getProjectByEnv(),
           getEnvironmentListValid(projectId),
           deployValid(),
         ],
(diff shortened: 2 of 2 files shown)


test/commands/deploy.test.js
```diff
@@ -63,9 +63,8 @@ describe('deploy', () => {
       it('should not display the list of projects', () => testCli({
         env: testEnv2,
         token: 'any',
-        command: () => DeployCommand.run([]),
+        command: () => DeployCommand.run(['-p', '82']),
         api: [
-          getProjectByEnv(),
```
rap2hpoutre, line 68: There might be actually two cases that we want to test. Before this addition, the test `when project was provided it should not display the list of projects` expects that project prompt is not shown when you use it in an environment.

Suggestion (tested locally):
 - when project has been guessed via environment it should not display the list of projects _(e.g: the test that exists currently)_
 - when project was provided via parameters it should not display the list of projects  _(e.g: the test you added)_
```js
    describe('when project has been guessed via environment', () => {
      const projectId = 82;
      const environmentName = 'name1';
      it('should not display the list of projects', () => testCli({
        env: testEnv2,
        token: 'any',
        command: () => DeployCommand.run(['-p', '82']),
        api: [
          getEnvironmentListValid(projectId),
          deployValid(),
        ],
        std: [
          ...inOutSelectEnvironment(environmentName),
          ...inOutConfirmDeploy(environmentName),
        ],
      }));
    });

    describe('when project was provided via parameters', () => {
      const projectId = 82;
      const environmentName = 'name1';
      it('should not display the list of projects', () => testCli({
        env: testEnv2,
        token: 'any',
        command: () => DeployCommand.run(['-p', '82']),
        api: [
          getEnvironmentListValid(projectId),
          deployValid(),
        ],
        std: [
          ...inOutSelectEnvironment(environmentName),
          ...inOutConfirmDeploy(environmentName),
        ],
      }));
    });
```

By doing this two tests (i.e: keep the original one and add a new one that validate this feat), we are sure there is no regression AND the new feat works. WDYT?

src/commands/deploy.js
```diff
@@ -112,6 +112,11 @@ DeployCommand.flags = {
     char: 'f',
     description: 'Skip deploy confirmation.',
   }),
+  projectId: flags.string({
+    char: 'p',
+    description: 'Forest project ID.',
```
rap2hpoutre, line 117: Should we validate the description with the product team?

In `init` command, the text is:

> The id of the project you want to init.

`push` command:

> The id of the project to work on.


So it could be: 

```suggestion
    description: 'The id of the project you want to deploy.',
```

(My suggestion is not that good I guess. Maybe just: `The id of the project to work on`)

Anyway `create` and environments commands use "Forest project ID"...

As you wish then!

rap2hpoutre (changes requested):
See the two comments below (the only one that is important is the one about tests). Maybe we should rename the PR with `feat` instead of `fix` (everything works just like before, but we _"allow project to be passed in as argument"_.

SteveBunlon:
> See the two comments below (the only one that is important is the one about tests). Maybe we should rename the PR with `feat` instead of `fix` (everything works just like before, but we _"allow project to be passed in as argument"_. I may be wrong though.

It is defined as a bug in clickup, so I would rather prefer to keep it as a bug 🙏

src/commands/deploy.js
SteveBunlon: I don't think this should be taken into a whole process of validation. But indeed your suggestion is way better than what I did 👍

test/commands/deploy.test.js
SteveBunlon: Actually, I though about this but, I don't think it's worth it because the test suite has been developed in such a way that forces the developper to specify the projectId even if you pass the projectId in. (see the following: `getEnvironmentListValid(projectId)`)

So according to me, it's just adding two different labels in two distinct `describe`, but with the same code. 

According to me the two options (mine and yours) are both correct and incorrect on some points. 

I would love to have your final thoughts on this 🙏

rap2hpoutre (approved):
🚀

PR #123, opened 2020-11-20
fix(push-command): allow project to be passed in as argument using --projectId or -p

SteveBunlon:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -78,14 +78,16 @@ PushCommand.flags = {
     char: 'e',
     description: 'The remote environment name to push onto.',
   }),
-  projectId: flags.string({
-    description: 'The id of the project to work on.',
-  }),
   force: flags.boolean({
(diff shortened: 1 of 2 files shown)


src/commands/push.js
```diff
@@ -78,15 +78,17 @@ PushCommand.flags = {
...
     description: 'Skip push changes confirmation.',
   }),
   help: flags.boolean({
     description: 'Display usage information.',
   }),
+  projectId: flags.string({
+    char: 'p',
+    description: 'The id of the project to work on.',
+    default: null,
+  }),
```
arnaud-moncel, line 91: Why do you move this ?

arnaud-moncel (approved):
Good job 👍 👑

src/commands/push.js
SteveBunlon: To ease copy past while debugging

PR #127, opened 2020-12-03
fix: prevent installation error due to premature evaluation caused by a dependency

### LINKED ISSUE #126: Broken dependency - cli-table
Hey guys,
first of all - kudos on a great product!

Opening an issue to point out that for the past couple of hours, running `npm i forest-cli` crashes, saying "`premature evluation`".

Through the logs, we narrowed down the issue to a new release of [cli-table](https://github.com/Automattic/cli-table/releases) that seems to be broken in itself.

This small issue is causing the cascading effect of breaking our entire CI/CD chain, and we would love to see if you can get it fixed on your side. 😅 (Even though it seems like a problem that arised completely unrelated to you)

Tagging my lovely colleagues: @eyalhakim @gabimor @nuritlh @robozevel @odeds

yotamofek:
## Pull Request checklist:

- [ ] Create automatic tests
- [ ] Review my own code (syntax, style, simplicity, readability)

Fixes #126 by specifying commit hash explicitly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@oclif/plugin-warn-if-update-available": "1.7.0",
     "bluebird": "3.5.2",
     "chalk": "4.0.0",
-    "cli-table": "github:Automattic/cli-table#master",
+    "cli-table": "github:Automattic/cli-table#7b14232ba779929e1859b267bf753c150d90a618",
     "clipboardy": "2.3.0",
     "dotenv": "8.2.0",
     "inquirer": "6.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1986,7 +1986,7 @@ cli-table3@^0.5.0, cli-table3@^0.5.1:
   optionalDependencies:
     colors "^1.1.2"
 
-cli-table@^0.3.1, "cli-table@github:Automattic/cli-table#master":
+cli-table@^0.3.1, "cli-table@github:Automattic/cli-table#7b14232ba779929e1859b267bf753c150d90a618":
   version "0.3.1"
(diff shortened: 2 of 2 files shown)


ghusse:
Hello @yotamofek, thanks a lot for your detailed diagnostic on this issue, and your fast solution proposal!

👍

package.json
```diff
@@ -23,7 +23,7 @@
     "@oclif/plugin-warn-if-update-available": "1.7.0",
     "bluebird": "3.5.2",
     "chalk": "4.0.0",
-    "cli-table": "github:Automattic/cli-table#master",
+    "cli-table": "github:Automattic/cli-table#7b14232ba779929e1859b267bf753c150d90a618",
```
ghusse, line 26: For what I can see from their repo, they tagged a new version 0.3.3 recently, but did not release it. 

I suppose that this is the safer way to go for the moment.

yotamofek: Yeah, the release is very recent (around 2 hours ago) and it seems that it failed to build. The commit hash is the last one that worked on an `npm` project I had that already had `cli-table` installed. Also, they seemed to change the pkg name in one of the latest commits, which also seems to break stuff.

Until there's an `npm install`-able version of `cli-table` (currently `npm i cli-table` doesn't work either), I definitely think that this is the safest approach. A more holistic fix, IMHO, would be to not point to the package through github, once that's possible.

arnaudbesnier (approved):

ghusse (approved):
Thanks a lot for your PR!

I changed the title to allow it it be released as a fix, once merged.

PR #160, opened 2021-04-08
fix(dwo): allow to deploy on remote non production reference environment

GuillaumeCisco:
## Pull Request checklist:

- [ ] Create automatic tests
- [ ] Review my own code (syntax, style, simplicity, readability)

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -55,14 +55,7 @@ class DeployCommand extends AbstractAuthenticatedCommand {
    * @see getEnvironment
    */
   async selectEnvironment(environments) {
-    const choices = environments
-      .reduce((accumulator, environment) => {
-        if (environment.type === 'production' || (environment.currentBranch && environment.currentBranch.origin === null)) {
-          return accumulator;
-        }
-        accumulator.push({ name: environment.name });
-        return accumulator;
-      }, []);
+    const choices = environments.map(({ name }) => name);
     const response = await this.inquirer.prompt([{
       name: 'environment',
       message: 'Select the environment containing the layout changes you want to deploy to the reference environment',
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -34,18 +34,7 @@ class DeployCommand extends AbstractAuthenticatedCommand {
    */
   async getEnvironment(config) {
     const environments = await new EnvironmentManager(config).listEnvironments();
-    const productionExists = environments.find((environment) => environment.type === 'production');
-
-    if (environments.length === 0) throw new Error('❌ No environment found.');
-    if (!productionExists) {
-      const remoteEnvironments = environments.filter((environment) => environment.type === 'remote');
-      if (!remoteEnvironments.length) {
-        throw new Error('❌ No reference environment, please create a remote or production environment.');
-      }
-    }
-
-    const environmentName = config.ENVIRONMENT_NAME
-      || await this.selectEnvironment(environments);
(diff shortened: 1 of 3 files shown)


src/commands/deploy.js
```diff
@@ -29,15 +29,20 @@ class DeployCommand extends AbstractAuthenticatedCommand {
    * Get selected environment; prompt for environment when none is provided.
    * @param {Object} config - Actual command config (including command parameters).
    * @throws Will throw an error when there is no environment (unreachable).
-   * @throws Will throw an error when there is no production environment.
+   * @throws Will throw an error when there is no reference environment.
    * @return {Object} The environment found.
    */
   async getEnvironment(config) {
     const environments = await new EnvironmentManager(config).listEnvironments();
     const productionExists = environments.find((environment) => environment.type === 'production');
```
slimee, line 37: `productionExists` have no place anymore here
I think you can drop entirely this section, so getEnvironment will be:

`const environmentName = config.ENVIRONMENT_NAME
      || await this.selectEnvironment(environments);
    return environments.find((environment) => environment.name === environmentName);`

slimee: what do you this of that?

slimee: We can discuss how to enhance `listEnvironments` to exclude the reference in another scope

slimee: IMO I will split `listEnvironments` into:
 - `listDeployableEnvironments`  for command deploy
 - `listEnvironmentToPushOnto` for command push
 - `listAllEnvironments` for environment:list command

GuillaumeCisco: I agree, will do the change

slimee (approved):
🚀 
Note: with this PR, the toolbelt gives the choice to deploy onto the reference which is invalid.
We should enhance the list of choice in another PR

src/commands/deploy.js
```diff
@@ -29,18 +29,12 @@ class DeployCommand extends AbstractAuthenticatedCommand {
...
    * @throws Will throw an error when there is no environment (unreachable).
-   * @throws Will throw an error when there is no production environment.
+   * @throws Will throw an error when there is no reference environment.
    * @return {Object} The environment found.
    */
   async getEnvironment(config) {
     const environments = await new EnvironmentManager(config).listEnvironments();
-    const productionExists = environments.find((environment) => environment.type === 'production');
-
-    if (environments.length === 0) throw new Error('❌ No environment found.');
```
slimee, line 39: this line should be kept I think
If the list is empty it's good to stop the action

src/commands/deploy.js
```diff
@@ -50,14 +44,12 @@ class DeployCommand extends AbstractAuthenticatedCommand {
...
     const response = await this.inquirer.prompt([{
       name: 'environment',
-      message: 'Select the environment containing the layout changes you want to deploy to production',
+      message: 'Select the environment containing the layout changes you want to deploy to the reference environment',
       type: 'list',
-      choices: environments
-        // NOTICE: Remove production since it should not be deployable on itself.
-        .filter((environment) => environment.type !== 'production')
-        .map(({ name }) => name),
+      choices,
```
slimee, line 52: I think this should be kept as is.
We should remove the reference, since we are not able to do this now, it's better to remove the production instead of removing nothing.

GuillaumeCisco: ok will rollback ;)

slimee (approved):
🚀

PR #170, opened 2021-05-20
feat(switch-command): add branch:switch alias

SteveBunlon:
## Pull Request checklist:

- [ ] Create automatic tests

It is actually impossible to test because of the test format we currently have

I couldn't decide whether to alias the command with `forest branch:switch` or to make the command to actually be `forest branch:switch` and add `forest switch` as an alias. What would you suggest ? My bet is that other commands related to DWO don't use the entity:action pattern, so let's keep it as it is and add the `forest branch:switch` as an alias.


guillaumedeslandes (approved):
🚀

PR #219, opened 2021-09-01
feat: display the auth secret when creating an environment

slimee:
### General

### Security

### DIFF
--- a/src/services/environment-manager.js
+++ b/src/services/environment-manager.js
@@ -8,8 +8,10 @@ const DeploymentRequestSerializer = require('../serializers/deployment-request')
 const JobStateChecker = require('../services/job-state-checker');
 
 function EnvironmentManager(config) {
-  const { assertPresent, authenticator, env } = Context.inject();
-  assertPresent({ authenticator, env });
+  const {
+    assertPresent, authenticator, env, keyGenerator,
+  } = Context.inject();
+  assertPresent({ authenticator, env, keyGenerator });
 
   this.listEnvironments = async () => {
     const authToken = authenticator.getAuthToken();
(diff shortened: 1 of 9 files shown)


test/commands/test-cli-helper/test-cli-context.js
```diff
@@ -71,6 +71,7 @@ const preparePlan = ({
   env,
   prompts,
   tokenBehavior,
+  additionnalStep,
```
slimee, line 74: the change in this file is to support the `additionnalStep` keyword, which adds a step at the end of the plan.

test/commands/test-cli-helper/test-cli-errors.js
```diff
@@ -59,7 +59,9 @@ function errorIfStdRest(stds) {
   }
 }
 
-function errorIfBadCommand({ commandClass, commandArgs, commandPlan }) {
+function errorIfBadCommand({
```
slimee, line 62: the diff in this file is to support the 'additionnalStep' keyword, which adds a step at the end of the plan.

jeffladiray (commented):
LGTM, but 
image
`FOREST_AUTH_SECRET` seems a bit long compared to the one generated on the frontend. 

That's not an issue, but for consistency, I would suggest to reduce the length to 48 (Just like the frontend does)

src/services/environment-manager.js
```diff
@@ -36,7 +38,7 @@ function EnvironmentManager(config) {
   this.createEnvironment = async () => {
     const authToken = authenticator.getAuthToken();
 
-    return agent
+    const environment = await agent
```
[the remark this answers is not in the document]
slimee: why not!

jeffladiray (approved):
Look good, works as expected.

Feel free to ignore my comment if you don't find it useful, maybe that's just me :D

PR #280, opened 2021-11-03
chore: add TO on workflows

RomainCoudour:
With the latest 'bricolage', some steps can be infinite. One reason is: the step cannot find a host to start.. (is it even related ?)
To prevent this issue to become an economic wormhole, timeout-minutes is added

Issue1: https://github.com/ForestAdmin/toolbelt/actions/runs/1417190894
Issue2: https://github.com/ForestAdmin/toolbelt/runs/4091624674?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Cancel previous running workflows
         uses: fkirc/skip-duplicate-actions@9da67cec51d21092667038f0f4fda73099a3a451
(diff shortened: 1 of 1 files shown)


adriguy (commented):
To me, this could not be a good solution 🤔  
Should we rather find the root cause of those "infinite" steps?

I know this project becomes very time-consuming (maybe frustrating) but I am afraid this solution will backfire in the near future (developers encountering timeouts on their CI tests would obviously complain)...

RomainCoudour:
Potential issue: https://www.githubstatus.com/incidents/4cnqjxrv0vzg
w8ing for a couple more CI runs to complete in order to close this issue

RomainCoudour:
The issue seems to be related to GitHub Actions' degraded service for Nov, 3rd. Nothing more to do on this side.

PR #316, opened 2021-12-22
chore: upgrade context

slimee:
context lib was updated, this leads to few changes in context construction


guillaumedeslandes (approved):
No obvious error from here.

PR #320, opened 2021-12-30
refactor(logger): allow to change the message color

Scra3:
### General

### Security

### DIFF
--- a/src/services/logger.js
+++ b/src/services/logger.js
@@ -33,7 +33,7 @@ class Logger {
 
     let actualMessage;
     if (options.colorLine) {
-      actualMessage = `${actualPrefix}${chalk.bold[options.colorLine](message)} \n`;
+      actualMessage = `${actualPrefix}${chalk[options.colorLine](message)} \n`;
     } else {
       actualMessage = `${actualPrefix}${message} \n`;
     }
--- a/src/services/logger.js
+++ b/src/services/logger.js
@@ -45,15 +45,15 @@ class Logger {
     }
   }
 
-  _logLines(messagesWithPotentialGivenOption, baseOptions) {
-    const { options, messages } = Logger._extractGivenOptionFromMessages(
-      messagesWithPotentialGivenOption,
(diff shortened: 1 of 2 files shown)


src/services/logger.js
```diff
@@ -23,14 +23,20 @@ class Logger {
...
     if ([undefined, null, ''].indexOf(options.prefix) === -1) actualPrefix = `${options.prefix} `;
     if (actualPrefix && options.color) actualPrefix = chalk.bold[options.color](actualPrefix);
 
-    const actualMessage = `${actualPrefix}${message} \n`;
+    let actualMessage;
+    if (options.lineColor) {
+      actualMessage = `${actualPrefix}${chalk[options.lineColor](message)} \n`;
+    } else {
+      actualMessage = `${actualPrefix}${message} \n`;
+    }
```
guillaumedeslandes, line 39:
```suggestion
    let actualMessage = message;
    if (options.lineColor) {
      actualMessage = `${chalk[options.lineColor](message)}`;
    }
    actualMessage = `${actualPrefix}${actualMessage}\n`;
```

Don't know why we let an space before the new line character 🤷‍♂️ Removing it does not seem to impact the test suite.

src/services/logger.js
```diff
@@ -39,8 +45,22 @@ class Logger {
...
+    const { options, messages } = Logger._extractGivenOptionFromMessages(
+      messagesWithPotentialGivenOption,
+    );
+    messages.forEach((message) => this._logLine(message, { ...baseOptions, ...options }));
+  }
+
+  // this method is a hack to keep the signature of all existing public methods.
+  static _extractGivenOptionFromMessages(messages) {
+    let options = {};
+    const potentialGivenOptionFromCommand = messages[messages.length - 1];
```
guillaumedeslandes, line 58: This is too permissive.
Any object given as the last parameter will be considered as an option object.
In your case, checking there is only one key ('lineColor`) should be done.

guillaumedeslandes (changes requested):
See comments 🙏

guillaumedeslandes (approved):

PR #360, opened 2022-03-23
fix(datatypes): MONEY is not automatically generated anymore on Postgresql and MsSql

elsak02:
### General

### Security


elsak02:
As Sequelize does not support MONEY Datatype and Postgres has limitations and does not recommend its usage (rounding errors, limited storage size) (See https://www.postgresql.org/docs/14/datatype-money.html). It has been decided to not support it by FA and a log message will be sent through the CLI once this database type is used.

src/services/schema/update/analyzer/sequelize-column-type-getter.js
```diff
@@ -122,8 +122,6 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
       case 'DOUBLE':
       case 'DOUBLE PRECISION':
       case typeContains(type, 'DECIMAL'):
-      case 'MONEY': // MSSQL type
```
SteveBunlon, line 125: It would be super great to have non-regression test on this. We are changing the behaviour there, and no tests are broken, meaning this section is not tested correctly.

SteveBunlon (changes requested):
🛡️ Let's add some little test on this 🛡️

SteveBunlon (approved):
🦶 🏈   🥅

SteveBunlon (approved):
🦶 🏈   🥅

PR #492, opened 2023-01-24
chore(typescript): add typescript to toolbelt

SteveBunlon:
### General

### Security

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,8 +85,6 @@ jobs:
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', '**/package.json') }}
       - name: Build
         run: yarn build
-      - name: Build
-        run: yarn build
       - name: Upload build
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,8 +76,8 @@ jobs:
     steps:
       - name: Cancel previous running workflows
         uses: fkirc/skip-duplicate-actions@master
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
           node-version: 16.14.0
       - uses: actions/cache@v3
@@ -90,7 +90,12 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: build
-          path: '/dist/*'
+          path: |
+            /dist/*
+            package.json
+            /bin
+            /npm-shrinkwrap.json
+            /oclif.manifest.json
           retention-days: 1
 
   deploy:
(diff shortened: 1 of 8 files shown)


SteveBunlon:
This PR also needs to update to github actions. They are incoming.

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
```
SteveBunlon, line 17: @arnaud-moncel @jeffladiray I'm pretty sure you could have useful insights / comments on what I've changed. Otherwise, if you could point me to someone helpful, that would be awesome ❤️

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
...
       - uses: actions/setup-node@v2
         with:
-          node-version: 14.17.6
+          node-version: 16.14.0
```
jeffladiray, line 20: Was this change needed?

No issue for me to keep it in this PR, but I'm wondering since we're migrating to node 18 if it shouldn't be included here

.github/workflows/build.yml
```diff
@@ -69,11 +69,35 @@ jobs:
       - name: Test postpack
         run: yarn postpack
 
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    needs: [test]
```
jeffladiray, line 75: Shouldn't we build and test in parallel, then trigger the release?

.github/workflows/build.yml
```diff
@@ -92,6 +116,10 @@ jobs:
         run: |
           VERSION=`node -p "require('./package.json').devDependencies['semantic-release']"`
           echo "::set-output name=version::$VERSION"
+      - name: Download build
```
jeffladiray, line 119: If we only need the artifact to publish, I think that `action/checkout@v2` can be remove ?

package.json
```diff
@@ -77,7 +79,9 @@
     "nock": "13.0.4",
     "rimraf": "3.0.2",
     "semantic-release": "19.0.3",
-    "semantic-release-slack-bot": "3.5.2"
+    "semantic-release-slack-bot": "3.5.2",
+    "ts-jest": "^27.1.2",
```
jeffladiray, line 83: Not sure if that's relevent here, but on the server we usually use `swc/jest` instead of `ts-jest`

.github/workflows/build.yml
SteveBunlon: You are totally right

SteveBunlon: done, now testing and building needs lint (for energy saving purposes) and release needs build and test (they are run in parralele)

.github/workflows/build.yml
SteveBunlon: Actually you are right, I'll remove this from this PR, and let's focus on node versions later on, this PR is primarily about typescript 👌

.github/workflows/build.yml
SteveBunlon: To be honest, I don't know. How is the release performed ? Will the action merge the whitelisted content from the github repo + the cached dist folder ? 

This is something I can test actually, can you remind me the steps to attempt a release on beta ?

SteveBunlon: Done

package.json
SteveBunlon: Done thanks to @Thenkei

PR #514, opened 2023-02-09
refactor(project-create): factorize code that will be reused

DrRaider:
#860puxe6q

### General

### Security


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3.4'
 services:
   MongoDB-MIN:
-    image: mongo:3.2
+    image: mongo:4.2
```
DrRaider, line 4:
```suggestion
    image: mongo:3.2
```

src/abstract-authenticated-command.ts
```diff
@@ -0,0 +1,43 @@
+import type * as Config from '@oclif/config';
+import Authenticator from './services/authenticator';
+import AbstractCommand from './abstract-command';
+
+export default abstract class AbstractAuthenticatedCommand extends AbstractCommand {
+  protected authenticator: Authenticator;
```
SteveBunlon, line 6:
```suggestion
  protected readonly authenticator: Authenticator;
```

src/abstract-project-create-command.ts
```diff
@@ -0,0 +1,271 @@
+import type {
```
romain-gilliotte, line 1: I don't understand the point of this file

- This is used only once, why split the projects/create?
- Why is that file not in `commands/project`

src/abstract-authenticated-command.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(argv: string[], config: Config.IConfig, plan?) {
+    super(argv, config, plan);
+
+    const { assertPresent, authenticator } = this.context;
+    assertPresent({ authenticator });
+
+    this.authenticator = authenticator;
+  }
+
+  async checkAuthentication() {
```
romain-gilliotte, line 18: should be private?

src/abstract-authenticated-command.ts
```diff
@@ -0,0 +1,55 @@
...
+      await this.authenticator.tryLogin({});
+      if (!this.authenticator.getAuthToken()) {
+        this.exit(10);
+      }
+    }
+  }
+
+  // Authentication error handler.
+  async handleAuthenticationErrors(error) {
+    // NOTICE: Due to ip-whitelist, 404 will never be thrown for a project
```
romain-gilliotte, line 30: should be private

src/abstract-authenticated-command.ts
romain-gilliotte: Return type?

src/abstract-authenticated-command.ts
```diff
@@ -0,0 +1,55 @@
+import type Authenticator from './services/authenticator';
+import type * as Config from '@oclif/config';
+
+import AbstractCommand from './abstract-command';
+
+export default abstract class AbstractAuthenticatedCommand extends AbstractCommand {
+  protected readonly authenticator: Authenticator;
```
romain-gilliotte, line 7: Difficult to read because the method ordering/visibility is not predictable (this, or I don't understand the ordering)

src/abstract-project-create-command.ts
```diff
@@ -0,0 +1,271 @@
...
+  AppConfig,
```
romain-gilliotte, line 2: Check visibility of method.

How do you decide what is public/private?

romain-gilliotte (approved):

src/abstract-project-create-command.ts
DrRaider: - It will be used in the new commands
- the commands folder is the one "scanned" by oclif, so the file would appear as an actual command

PR #515, opened 2023-02-13
ci(build): use the same cache hash between lint and build

DrRaider:
### General

### Security

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,9 +83,7 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: '**/node_modules'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -84,6 +84,8 @@ jobs:
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', '**/package.json') }}
+      - name: install dependencies
```
SteveBunlon, line 87: The issue resides in the node modules cache keys that is different between lint and build. We should be using the same cache key, my bad !

SteveBunlon (approved):

PR #517, opened 2023-02-17
feat(projects:create): add new command to generate nosql projects for agent-nodejs

DrRaider:
https://app.clickup.com/t/860puxe73
### General

### Security

### DIFF
--- a/src/services/schema/update/database.js
+++ b/src/services/schema/update/database.js
@@ -70,7 +70,10 @@ class Database {
       if (options.dbPassword) {
         connectionUrl += `:${options.dbPassword}`;
       }
-      connectionUrl += `@${options.dbHostname}`;
+      if (options.dbUser || options.dbPassword) {
+        connectionUrl += '@';
+      }
+      connectionUrl += options.dbHostname;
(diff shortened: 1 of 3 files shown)


src/services/schema/update/database.js
```diff
@@ -70,7 +70,10 @@ class Database {
       if (options.dbPassword) {
         connectionUrl += `:${options.dbPassword}`;
       }
-      connectionUrl += `@${options.dbHostname}`;
+      if (options.dbUser || options.dbPassword) {
+        connectionUrl += '@';
+      }
+      connectionUrl += options.dbHostname;
```
DrRaider, line 76: Before changing this, we couldn't connect to a database without credentials using the test helper.
Now we are ISO with the documentation: https://www.mongodb.com/docs/manual/reference/connection-string/#standard-connection-string-format

DrRaider: The helper was working for other tests since we never went through the connectionString computing on them.
Here if we pass the same URL `mongodb://localhost:27016` it was transformed to `mongodb://@localhost:27016` which the driver does not accept.

PR #523, opened 2023-02-21
ci: fix build

DrRaider:
### General

### Security

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,10 +57,6 @@ jobs:
           path: "**/node_modules"
           key: ${{ runner.os }}-node_modules-${{ hashFiles('**/yarn.lock') }}
 
-      - name: Install dependencies
-        if: steps.cache-primes.outputs.cache-hit != 'true'
-        run: yarn install --frozen-lockfile --non-interactive --production=false
-
       - name: Build
         run: yarn build
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,22 +12,57 @@ jobs:
     steps:
       - name: Cancel previous running workflows
         uses: fkirc/skip-duplicate-actions@9da67cec51d21092667038f0f4fda73099a3a451
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -12,23 +12,58 @@ jobs:
...
+        uses: fkirc/skip-duplicate-actions@master
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14.17.6
+          cache: 'yarn'
+
+      - name: Cache node_modules
+        uses: actions/cache@v3
+        id: cache-primes
```
SteveBunlon, line 55: Some actions use this ID key, and some other does not, is it intended ?

SteveBunlon: I think I got it, this is an identifier to get the step and assert cache has been found. Correct ? If cache has been found based on yarn.lock hash, get it, otherwise install dependencies. It is a speed up measure.

SteveBunlon: That being said, knowing that any action requires the linter, I would say that force reinstalling the dependencies only in the lint step should be enough. WDYT ?

SteveBunlon (changes requested):

.github/workflows/build.yml
DrRaider: This is used to check we we fetched a cache or not, which allows to decide if we run `yarn install`.
The steps that use it are required for the "test" and "deploy" steps.
Since those two require either lint or build to start, we're sure to have a cache when they are triggered, so we don't execute `yarn install` and we don't need the key to check if we properly fetched a cache.

SteveBunlon: Ok following up my next message: 

> That being said, knowing that any action requires the linter, I would say that force reinstalling the dependencies only in the lint step should be enough. WDYT ?

This means that only the action at the root of anything needs to reinstall dependencies if not cache are present. So we can remove the `yarn install` from the `build` action correct ? (every actions are based on the `lint` one)

DrRaider: 👍

SteveBunlon (approved):

PR #524, opened 2023-02-21
refactor(dumper): create abstract dumper to regroup functionalities

SteveBunlon:
### General

### Security


src/interfaces/project-create-interface.ts
```diff
@@ -15,6 +15,10 @@ export interface AppConfig {
   applicationName: string;
   appHostname: string;
   appPort: number;
+  path?: string;
```
SteveBunlon, line 18: Don't have any clue where this could come from, to be investigated

src/services/dumper/abstract-dumper.ts
```diff
@@ -0,0 +1,85 @@
+import type { ConfigInterface } from '../../interfaces/project-create-interface';
```
SteveBunlon, line 1: This file needs appropriate unit tests

src/services/dumper/dumper.js
```diff
@@ -135,93 +91,82 @@ class Dumper {
...
-    this.writeFile(projectPath, 'package.json', `${JSON.stringify(pkg, null, 2)}\n`);
+    this.writeFile('package.json', `${JSON.stringify(pkg, null, 2)}\n`);
   }
 
   static tableToFilename(table) {
     return _.kebabCase(table);
   }
 
-  static getDatabaseUrl(config) {
+  static getDatabaseUrl(dbConfig) {
```
SteveBunlon, line 101: This could be in a utils function, and moved to the parent level as both of the dumpers will require it.

src/services/dumper/dumper.js
```diff
@@ -135,93 +91,82 @@ class Dumper {
...
-      connectionString = `${protocol}://${config.dbUser}${password}@${config.dbHostname}${port}/${config.dbName}`;
+      connectionString = `${protocol}://${dbConfig.dbUser}${password}@${dbConfig.dbHostname}${port}/${dbConfig.dbName}`;
     }
 
     return connectionString;
   }
 
-  static isDatabaseLocal(config) {
-    const databaseUrl = Dumper.getDatabaseUrl(config);
+  static isDatabaseLocal(dbConfig) {
```
SteveBunlon, line 127: This could be in a utils function, and moved to the parent level as both of the dumpers will require it.

src/services/dumper/dumper.js
SteveBunlon: Done

src/services/dumper/dumper.js
SteveBunlon: Actually nop

src/services/dumper/abstract-dumper.ts
SteveBunlon: Tests added

src/interfaces/project-create-interface.ts
SteveBunlon: Ok I think I got it, this has been added in the past to help integration of tests

SteveBunlon: Final conclusion: this has been added in the past to help implementation of tests, never change your code base to ease tests, I've removed it.

src/services/dumpers/dumper-v1.js
```diff
@@ -0,0 +1,378 @@
+const _ = require('lodash');
+const { URL } = require('url');
+const { plural, singular } = require('pluralize');
+const stringUtils = require('../../utils/strings');
+const toValidPackageName = require('../../utils/to-valid-package-name');
+const IncompatibleLianaForUpdateError = require('../../errors/dumper/incompatible-liana-for-update-error');
+const InvalidForestCLIProjectStructureError = require('../../errors/dumper/invalid-forest-cli-project-structure-error');
+require('../../utils/handlebars/loader');
```
SteveBunlon, line 8: This has nothing to do here. This should be loaded by the context [Optional]

SteveBunlon: Done moved to AbstractDumper which seems more consistent to me

src/services/dumpers/dumper-v1.js
```diff
@@ -0,0 +1,377 @@
+const _ = require('lodash');
```
SteveBunlon, line 1: Not a big fan of this name, if you have any suggestion, fire ! 

Something like agent-express-dumper.js ?

DrRaider: `forest-express-dumper` ? Since it's for `forest-express-sequelize` and `forest-express-mongoose`

src/utils/database-url.ts
```diff
@@ -0,0 +1,31 @@
+import type { DbConfigInterface } from '../interfaces/project-create-interface';
+
+export default function buildDatabaseUrl(config: DbConfigInterface) {
+  let connectionString;
```
romain-gilliotte, line 4: implicit any, nothing is typed here.

Can you set the return type of the function and type the `connectionString` variable?

src/utils/database-url.ts
```diff
@@ -0,0 +1,31 @@
...
+      protocol = 'mongodb+srv';
+      port = '';
+    }
+
+    if (config.dbPassword) {
+      // NOTICE: Encode password string in case of special chars.
+      password = `:${encodeURIComponent(config.dbPassword)}`;
+    }
+
+    connectionString = `${protocol}://${config.dbUser}${password}@${config.dbHostname}${port}/${config.dbName}`;
```
romain-gilliotte, line 27: It's less likely, but dbUser, dbHostname and dbName name also contain special chars.

Encoding URLs is natively supported in JS (with the URL class), we should really be using that instead of error-prone string manipulation.

```
const url = new URL('protocol://');
url.hostname = 'xxx'
url.port = String(2343)
url.username = 'romain';
url.password = '234234';
return url.href
```

test/services/analyzer/handlebars/helpers/wrap-quotes.unit.test.js
```diff
@@ -1,4 +1,4 @@
-const wrapQuotes = require('../../../../../src/services/dumper/handlebars/helpers/wrap-quotes');
+const wrapQuotes = require('../../../../../src/utils/handlebars/helpers/wrap-quotes');
```
romain-gilliotte, line 1: the source file has moved, you should move the test file to match

test/services/analyzer/schema-service.unit.test.js
```diff
@@ -8,7 +8,7 @@ const makeContext = () => ({
   },
```
romain-gilliotte, line 8: source file is in `services/schema/`
move the test

src/services/dumpers/dumper-v1.js
romain-gilliotte: Agreed
In the future this repo is very likely to support other agents (if we release standalone agents for all stacks)

I would call this `forest-express`, `agent-nodejs`, `forest-rails`, ...

src/interfaces/project-create-interface.ts
```diff
@@ -15,6 +15,9 @@ export interface AppConfig {
   applicationName: string;
   appHostname: string;
   appPort: number;
+  isUpdate?: boolean;
```
romain-gilliotte, line 18: Having an `isUpdate` field in a file which is named `project-create-interface` seems like a big hack.

src/commands/init.js
```diff
@@ -91,7 +92,19 @@ class InitCommand extends AbstractAuthenticatedCommand {
         const databaseConfiguration = await handleDatabaseConfiguration();
         this.spinner.continue();
         if (databaseConfiguration) {
-          this.environmentVariables.databaseUrl = buildDatabaseUrl(databaseConfiguration);
+          const dbConfig = {
```
romain-gilliotte, line 95: Why rename here, and not in the database-prompts.js file?

Is something preventing us from using the same name everywhere?

romain-gilliotte (changes requested):

src/services/dumpers/dumper-v1.js
DrRaider: Will do in the last PR, so that the rebasing of other PRs doesn't become hell

test/services/analyzer/handlebars/helpers/wrap-quotes.unit.test.js
DrRaider: 👍

src/commands/init.js
DrRaider: This is an even bigger refactor, which ultimately should be done, but we decided it was outside the scope here.

src/utils/database-url.ts
DrRaider: 👍

romain-gilliotte (approved):

src/commands/init.js
SteveBunlon: Totally agree with you @romain-gilliotte, but with @DrRaider as-well.

DrRaider (approved):

PR #534, opened 2023-02-27
chore(release-config): use a different place for the force-release tag

RomainCoudour:
Instead of losing the scope of the commit with a CI-related instruction, move this instruction to the end of the commit, surrounded by squared brackets (like the `skip ci` GH Action instruction)

Example of usage : `type(scope): subject [force release]`

Sources:
- https://github.com/semantic-release/commit-analyzer#configuration
- https://github.com/semantic-release/commit-analyzer#release-rules
- https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular
- https://github.com/micromatch/micromatch#matching-features

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -2,14 +2,14 @@ module.exports = {
   branches: ['main', { name:'beta', prerelease: true }],
   plugins: [
     [
-      '@semantic-release/commit-analyzer', {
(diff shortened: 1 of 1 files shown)


slimee (approved):
🚀

PR #537, opened 2023-03-06
feat(dumper): dump js mongodb models

SteveBunlon:
### General

### Security


SteveBunlon:
This PR contains additional works that needs to be taken out: 
- move lodash in the context
- strings utils is now in ts format inject in the context

This two PRs should be created after [this one](https://github.com/ForestAdmin/toolbelt/pull/524) has been merged

src/services/dumpers/agent-nodejs.ts
```diff
@@ -61,6 +67,7 @@ export default class AgentNodeJs extends AbstractDumper {
 
     if (dbDialect === 'mongodb') {
       dependencies['@forestadmin/datasource-mongoose'] = '^1.0.0';
+      dependencies.mongoose = '^6.8.3';
```
jeffladiray, line 70: Same as the previous one I guess, maybe we could upgrade to the latest 6.x ?

src/services/dumpers/agent-nodejs.ts
```diff
@@ -108,19 +115,23 @@ export default class AgentNodeJs extends AbstractDumper {
...
-      datasourceCreation: isMongoose
-        ? 'createMongooseDataSource(connection, {})'
-        : `
+      datasourceImport: null,
+      datasourceCreation: null,
+    };
+
+    if (isMongoose) {
+      context.datasourceImport = `const { createMongooseDataSource } = require('@forestadmin/datasource-mongoose');\nconst primaryConnection = require('./models/primary');`;
+      context.datasourceCreation = `createMongooseDataSource(primaryConnection, { flattenMode: 'auto' })`;
```
jeffladiray, line 124: So, models will be written in `./models/primary`? Any reason for the `primary`naming ?

src/services/dumpers/agent-nodejs.ts
```diff
@@ -191,7 +202,38 @@ export default class AgentNodeJs extends AbstractDumper {
...
+    });
+  }
+
+  private async writeMongooseModels(schema) {
+    await this.mkdirp(`${this.projectPath}/models/primary`);
+
+    this.copyHandleBarsTemplate('models/index.hbs', 'models/primary/index.js');
+
+    this.writeModels(schema);
+  }
```
jeffladiray, line 234: I think we should merge both function.

`writeModels` is mongoose specific atm, and `writeMongooseModels` does almost nothing outside creating the folder and copy index.

src/services/dumpers/agent-nodejs.ts
```diff
@@ -205,5 +247,9 @@ export default class AgentNodeJs extends AbstractDumper {
     this.writeDockerignore();
     this.writeDockerfile();
     this.writeDockerCompose(dumpConfig);
+
+    if (schema) {
```
jeffladiray, line 251:
```suggestion
    if (isMongoose) {
```
(I didn't test obviously, but this if(schema) -> writeMongoose seems weird to me)

jeffladiray (commented):
Same here, didn't review the tests.

These are pretty hard to read, I prefer to manually check that everyone of them fails if I try to modify the expected result.

src/services/dumpers/agent-nodejs.ts
SteveBunlon: Nop. I've just added the structure to add multi-database support by default. Primary is basically a name, could be `default`, `first-one`, even your name if you want.

src/services/dumpers/agent-nodejs.ts
SteveBunlon: Ok let's go.

src/services/dumpers/agent-nodejs.ts
SteveBunlon: I update bot the if statement and the schema variable name.

PR #552, opened 2023-03-14
refactor(arguments): allow to specify which arguments should be used for every project creation command

DrRaider:
### General

### Security

### DIFF
--- a/src/abstract-project-create-command.ts
+++ b/src/abstract-project-create-command.ts
@@ -1,5 +1,5 @@
+import type { createCommandArguments } from './interfaces/command-create-project-arguments-interface';
 import type { AppConfig, Config, DbConfig } from './interfaces/project-create-interface';
-import type CommandGenerateConfigGetter from './services/projects/create/command-generate-config-getter';
 import type { ProjectMeta } from './services/projects/create/project-creator';
 import type ProjectCreator from './services/projects/create/project-creator';
 import type Database from './services/schema/update/database';
@@ -7,16 +7,14 @@ import type Spinner from './services/spinner';
 import type EventSender from './utils/event-sender';
 import type Messages from './utils/messages';
 import type * as OclifConfig from '@oclif/config';
+import type { Input } from '@oclif/parser';
 
 import { flags } from '@oclif/command';
 
 import AbstractAuthenticatedCommand from './abstract-authenticated-command';
-import { dbDialectOptions } from './services/prompter/database-prompts';
 
 export default abstract class AbstractProjectCreateCommand extends AbstractAuthenticatedCommand {
   private readonly eventSender: EventSender;
-
-  private readonly commandGenerateConfigGetter: CommandGenerateConfigGetter;
 
   private readonly projectCreator: ProjectCreator;
 
@@ -26,7 +24,8 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
 
   protected readonly spinner: Spinner;
 
-  // Flags, args and Description must be defined on the class itself otherwise it cannot be parsed properly
+  protected abstract readonly agent: string | null;
+
   static override flags = {
     applicationHost: flags.string({
       char: 'H',
@@ -47,14 +46,6 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
       dependsOn: [],
       description: 'Enter the database credentials with a connection URL.',
       exclusive: ['ssl'],
-      required: false,
-    }),
-    databaseDialect: flags.string({
-      char: 'd',
-      dependsOn: [],
-      description: 'Enter your database dialect.',
-      exclusive: ['databaseConnectionURL'],
-      options: dbDialectOptions.map(option => option.value),
       required: false,
     }),
     databaseName: flags.string({
@@ -91,24 +82,11 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
       exclusive: ['databaseConnectionURL'],
       required: false,
     }),
-    databaseSchema: flags.string({
-      char: 's',
-      dependsOn: [],
-      description: 'Enter your database schema.',
-      exclusive: [],
-      required: false,
-    }),
     databaseSSL: flags.boolean({
       default: false,
       dependsOn: [],
       description: 'Use SSL for database connection.',
       exclusive: [],
-      required: false,
-    }),
-    mongoDBSRV: flags.boolean({
-      dependsOn: [],
-      description: 'Use SRV DNS record for mongoDB connection.',
-      exclusive: ['databaseConnectionURL'],
       required: false,
     }),
   };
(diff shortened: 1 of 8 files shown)


src/abstract-project-create-command.ts
```diff
@@ -140,69 +119,119 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
...
+        dbConfig,
+        appConfig,
+        forestAuthSecret: authSecret as string,
+        forestEnvSecret: envSecret as string,
+      });
+
+      await this.notifySuccess();
+    } catch (error) {
+      // Display customized error for non-authentication errors.
+      if (error.status !== 401 && error.status !== 403) {
```
SteveBunlon, line 163: Knowing we are in a `runAuthenticated` method, checking if the error is a 401 one is incorrect to me. At this stage, we are absolutely that the user is authenticated. So only 403 (authorization error) should and can occure here.

src/commands/projects/create.ts
```diff
@@ -1,34 +1,96 @@
...
   private readonly databaseAnalyzer: DatabaseAnalyzer;
 
   private readonly dumper: ForestExpress;
 
+  private readonly commandGenerateConfigGetter: CommandGenerateConfigGetter;
+
+  private readonly _agent: string | null = null;
+
+  static override readonly flags = {
+    ...AbstractProjectCreateCommand.flags,
```
SteveBunlon, line 29: This feels strange to be honest. On the mother class, it is not specified that flags should be overridden in any place. So, no contracts are specified anywhere that flags should be on the child class. This means that you will rely on the developper comprehension, where OOP is all about specifying contracts that developper should respect. Plus, the child class is responsible for getting back the mother flags. In OOP, the mother class should have this ownership.

src/abstract-authenticated-command.ts
```diff
@@ -15,7 +15,18 @@ export default abstract class AbstractAuthenticatedCommand extends AbstractComma
     this.authenticator = authenticator;
   }
 
-  async checkAuthentication() {
+  async run(): Promise<void> {
```
SteveBunlon, line 18: Is this really related to the flags?

src/abstract-project-create-command.ts
```diff
@@ -140,69 +119,119 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
...
+        this.logger.error(['Cannot generate your project.', `${this.messages.ERROR_UNEXPECTED}`]);
+        this.logger.log(`${this.chalk.red(error)}`);
+        this.exit(1);
+      } else {
+        throw error;
+      }
+    }
+  }
+
+  protected abstract processArguments(programArguments: { [name: string]: any }): Promise<{
```
SteveBunlon, line 173: This function name is not self explanatory...

I would rather call it getConfigFormArguments

src/interfaces/project-create-interface.ts
```diff
@@ -40,3 +40,38 @@ export interface Config {
   forestAuthSecret: string;
   forestEnvSecret: string;
 }
+
+interface DbProcessedArgumentsBase {
```
SteveBunlon, line 44: Again, processed does not mean anything to me. This looks like command arguments, so this should be called something like `createCommandDbArguments`.

All the stuff you added here is not related to project creation, but command argument parsing. This should reside in a different file, something like `command-create-project-arguments`

src/services/dumpers/forest-express.js
```diff
@@ -46,7 +46,6 @@ class ForestExpress extends AbstractDumper {
     this.isDatabaseLocal = isDatabaseLocal;
     this.toValidPackageName = toValidPackageName;
     this.strings = strings;
-    this.lodash = lodash;
```
SteveBunlon, line 49: ??

src/utils/agents.ts
```diff
@@ -0,0 +1,7 @@
+enum Agents {
+  NodeJS = 'agent-nodejs',
+  ExpressSequelize = 'express-sequelize',
+  ExpressMongoose = 'express-mongoose',
+}
+
+export default Agents;
```
SteveBunlon, line 7:
```suggestion
export default enum Agents {
  NodeJS = 'agent-nodejs',
  ExpressSequelize = 'express-sequelize',
  ExpressMongoose = 'express-mongoose',
}
```

SteveBunlon: Plus, this has nothing to do with flags

src/services/prompter/database-prompts.js
```diff
@@ -25,17 +26,17 @@ class DatabasePrompts extends AbstractPrompter {
...
+    this.handleDialect(forSql, forNosql);
     this.handleName();
     this.handleSchema();
     this.handleHostname();
     this.handlePort();
     this.handleUser();
     this.handlePassword();
     this.handleSsl();
-    this.handleMongodbSrv();
+    if (forNosql) this.handleMongodbSrv();
```
SteveBunlon, line 39: We already have this kind of checks. What is requested in prompts is handled by the `requests` object. Why didn't you use it to specify the prompts ?

SteveBunlon (changes requested):
I would split it in two. One PR to reorganise the args necessary for a command.

One PR which, based on the new granularity of args, can specify wether or not some options should be requested or not for a command

src/abstract-project-create-command.ts
DrRaider: The token of the user can be discarded during the command execution.
I simply moved this code around, this is not new code.

src/commands/projects/create.ts
DrRaider: It's the only way to do it. Oclif inheritance is not good.

src/abstract-authenticated-command.ts
DrRaider: This is just some reordering, I can put it in another PR if you want.

src/services/prompter/database-prompts.js
DrRaider: Actually it's not realy handled by the `requests` object here, `mongoDbSrv` is always in the `requests` but on the prompt definition we only ask it if the dialect is mongo (so this if is useless you are right)

src/interfaces/project-create-interface.ts
DrRaider: 👍

src/abstract-project-create-command.ts
DrRaider: 👍

src/abstract-project-create-command.ts
SteveBunlon: 👌

src/commands/projects/create.ts
SteveBunlon: 👌

SteveBunlon (approved):

PR #575, opened 2023-03-28
feat(agent-nodejs): allow users to generate a project in typescript

DrRaider:
### General

### Security

### DIFF
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -32,7 +32,7 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
       char: 'l',
       description: 'In which language would you like to generate your sources?',
       options: languageList.map(language => language.name),
-      required: true,
+      required: false,
     }),
   };
 
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -32,8 +32,7 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
       char: 'l',
       description: 'In which language would you like to generate your sources?',
       options: Object.values(languages).map(language => language.name),
-      default: 'required',
-      required: false,
+      required: true,
     }),
   };
 
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -8,8 +8,8 @@ import type * as OclifConfig from '@oclif/config';
 import { flags } from '@oclif/command';
 
 import AbstractProjectCreateCommand from '../../../abstract-project-create-command';
-import { nosqlDbDialectOptions } from '../../../services/prompter/database-prompts';
 import Agents from '../../../utils/agents';
+import languages from '../../../utils/languages';
 
 export default class NosqlCommand extends AbstractProjectCreateCommand {
   private readonly dumper: AgentNodeJs;
@@ -28,15 +28,12 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
       exclusive: ['databaseConnectionURL'],
       required: false,
     }),
-    javascript: flags.boolean({
-      description: 'Generate your project in javascript.',
-      exclusive: ['typescript'],
-      default: context => !context.flags.typescript,
-    }),
-    typescript: flags.boolean({
-      description: 'Generate your project in typescript.',
-      exclusive: ['javascript'],
-      default: false,
+    language: flags.string({
+      char: 'l',
+      description: 'In which language would you like to generate your sources?',
+      options: Object.values(languages).map(language => language.name),
+      default: 'required',
+      required: false,
     }),
   };
 
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -9,7 +9,7 @@ import { flags } from '@oclif/command';
 
 import AbstractProjectCreateCommand from '../../../abstract-project-create-command';
 import Agents from '../../../utils/agents';
-import languages from '../../../utils/languages';
+import { languageList } from '../../../utils/languages';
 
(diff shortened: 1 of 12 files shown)


SteveBunlon:
This PR will be used to perform the tests on the typescript onboarding. It should be used in combinaison with https://github.com/ForestAdmin/forestadmin/pull/5934

End to End Tests (from UI to project runtime): 

Javascript (non regression tests): 

- [ ] Onboarding sql docker
- [ ] Onboarding sql npm
- [ ] Onboarding nosql docker
- [ ] Onboarding nosql npm

Typescript (non regression tests):
- [ ] Onboarding sql docker
- [ ] Onboarding sql npm
- [ ] Onboarding nosql docker
- [ ] Onboarding nosql npm

src/commands/projects/create/nosql.ts
```diff
@@ -28,6 +28,16 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
       exclusive: ['databaseConnectionURL'],
       required: false,
     }),
+    javascript: flags.boolean({
```
SteveBunlon, line 31: Quick question, should we really consider the javascript file ? If we opt for the following: 
- No flag is Javascript
- Typescript flag is for typescript

--> We could get ride of it. Wdyt ?

SteveBunlon: I think you did this to get the proposal from the prompts... I'm questioning the necessity of it.

DrRaider: yeah, but IMO having it in the command shows to the users that there are other language options, so I'd be in favor of having the option.

src/commands/projects/create/nosql.ts
```diff
@@ -28,6 +28,13 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
       exclusive: ['databaseConnectionURL'],
       required: false,
     }),
+    language: flags.string({
+      char: 'l',
+      description: 'In which language would you like to generate your sources?',
+      options: Object.values(languages).map(language => language.name),
+      default: 'required',
+      required: false,
```
SteveBunlon, line 36: [Question] What are these parameters used for ? I though this flag would have been required, accordingly to the prompts

src/commands/projects/create/sql.ts
```diff
@@ -34,6 +35,13 @@ export default class SqlCommand extends AbstractProjectCreateCommand {
       exclusive: [],
       required: false,
     }),
+    language: flags.string({
+      char: 'l',
+      description: 'Choose the language you want to use for your project.',
+      options: Object.values(languages).map(language => language.name),
+      default: 'required',
+      required: false,
```
SteveBunlon, line 43: [Question] What are these parameters used for ? I though this flag would have been required, accordingly to the prompts

SteveBunlon (changes requested):

test/services/prompter/application-prompts.test.js
```diff
@@ -177,6 +178,68 @@ describe('services > prompter > application prompts', () => {
...
         expect(env.applicationPort).not.toStrictEqual(FAKE_APP_PORT);
         expect(prompts).toHaveLength(0);
+        resetParams();
+      });
+    });
+  });
+
+  describe('handling language', () => {
+    describe('when the language option is not requested', () => {
+      it('should add a prompt to ask for it', () => {
```
SteveBunlon, line 188: Isn't it the opposite ? Not requesting the language should not prompt for it ?

SteveBunlon: I would test both cases, when the language is requested, and when it's not

src/commands/projects/create/sql.ts
DrRaider: The default is used to have a value defined if the user doesn't pass the argument. It allows us to know if the language is requested for the prompts.
The required is set to false because otherwise the flag must be provided when typing the command (which makes the prompt useless).

test/abstractProjectCreateCommand.unit.test.ts
```diff
@@ -103,6 +103,13 @@ describe('abstractProjectCreateCommand command', () => {
           exclusive: ['databaseConnectionURL'],
           required: false,
         }),
+        language: flags.string({
+          char: 'l',
+          description: 'Choose the language you wna tot use for your project.',
```
jeffladiray, line 108:
```suggestion
          description: 'Choose the language you want to use for your project.',
```

src/services/projects/create/command-generate-config-getter.js
```diff
@@ -18,24 +18,30 @@ const REQUESTS = {
...
     if (programArguments.databaseConnectionURL) {
-      return this.AVAILABLE_REQUESTS.forConnectionUrl;
+      requestList = REQUESTS.forConnectionUrl;
+    } else {
+      requestList = REQUESTS.forFullPrompt;
     }
-    return this.AVAILABLE_REQUESTS.forFullPrompt;
+
+    // This is the only way we can know if the command is for agent-nodejs project.
+    // For other commands, the language is not supported.
```
jeffladiray, line 39: Not sure to understand the comment nor the code below 😓

SteveBunlon:
⚠️ When we test the new TS onboarding + non regression on JS, please use both database config: one with SSL and one without SSL ⚠️

jeffladiray (commented):
- `npm build` command in dockerfile doesn't work. It should be `npm run build`, otherwise it just crash.
- `npm run build` doesn't work either, see screenshot
image
adding `skipLibCheck: true` fixes the issue.

- When deploying in production on heroku, it fails by default. `npm start` (The command that is run by default on heroku) uses nodemon, that is not installed (As it is a dev-dependency).

- Opening the project in visual code right after the onboarding already raises a few problems
image
It could be worth either disabling these checks or make sure they don't appear.

PR #591, opened 2023-05-10
fix(project-create): use @forestadmin/datasource-sql when generating sql projects

romain-gilliotte:
With this PR:
- validation is used on oclif params
- schema is not asked when the URL is provided
- sslmode is used instead of SSL for SQL projects
- testing the database connection is delegated to datasource-sql


jeffladiray (approved):
🚀

PR #672, opened 2024-03-08
chore: proposal to remove dep

Scra3:
### General

### Security


SteveBunlon (changes requested):
Actually ! JWT is also used in tests. By convention we should use libraries from context only. 

So, jwt should be added entirely in the context and this change should be propagated everywhere. That's quite some code !

PR #769, opened 2026-06-13
feat(environments): support --type and optional URL on environments:create

Gawtier:
## What

Lets `forest environments:create` create the **first production environment** headlessly:

- new `--type` flag (`production` | `remote`);
- `--url/-u` is now **optional**.

The request payload only includes `apiEndpoint` when a URL is given and `type` when provided; `projectId` keeps being sent via the `forest-project-id` header (IP-whitelist). Existing remote-environment creation is unchanged.

## Why / backend alignment

Verified against `forestadmin-server` (`environment-model.ts`, `environment-service.ts`):

- `EnvironmentType` = `development | production | remote`.
- **`apiEndpoint` is nullable for every type** (format/HTTPS validation only runs when a value is given) → making `--url` optional matches the server contract; an env without a URL is simply created **inactive** (`isActive = apimapVersionId && apiEndpoint`) until one is set.
- On creation, a missing `type` **defaults to `remote`** server-side; creating a `production` env when one already exists **returns the existing one** (idempotent); ordering rules (e.g. *"A production environment must be created before any other environment"*) are **enforced and messaged by the server** — the command relays `errors[0].detail`.
- `--type` is intentionally scoped to **`production | remote`**: development environments have their own command (`forest init` -> per-user `development-environment-for-user` endpoint), so exposing `development` here would be a confusing duplicate.

## Tests

- New command test: `--type production` with no URL.
- Existing `environments:create` suite unchanged and green.
- `yarn build` OK, full lint 0 errors, touched suites green.

Refs PRD-524.

### DIFF
--- a/src/commands/environments/create.js
+++ b/src/commands/environments/create.js
@@ -54,8 +54,12 @@ CreateCommand.flags = {
   }),
   url: Flags.string({
     char: 'u',
-    description: 'Application URL.',
-    required: true,
+    description: 'Application URL (optional for a production environment).',
+  }),
+  type: Flags.string({
+    char: 't',
+    description: 'Environment type. Use "production" to create the first production environment.',
+    options: ['production', 'remote'],
   }),
   format: Flags.string({
     char: 'format',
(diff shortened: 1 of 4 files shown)


Gawtier:
Superseded by #773 (intra-repo PR — CI now has the repo secrets).

PR #768, opened 2026-06-13
feat(signup): add forest signup command

Gawtier:
## What

Adds a headless `forest signup` command: creates a Forest Admin account (`POST /api/users`) then opens a session (`POST /api/sessions`), persisting the token in the same place as `forest login` (`~/.forest.d/.forestrc`). Flags: `--email`, `--password`, `--first-name`, `--last-name` (missing fields are prompted).

Mirrors the existing `login` structure (`command → authenticator → api`). Server errors are relayed verbatim — generic on an existing email (respecting the server's anti-enumeration), specific on validation (e.g. weak password).

## Why

First CLI primitive of the headless Standalone onboarding. Refs PRD-523.

## Tests

- Command-level: success, existing-email (409), invalid email, weak password (422).
- Unit (`authenticator`, `api`): `signup` / `trySignup` / `toSignupError` + `api.signup`.
- `yarn build` OK · full lint 0 errors · touched suites green (48 tests). DB-dependent integration suites need Mongo/Postgres (run in CI).

## Deferred / out of scope

- Compliance & marketing consent (ToS, `--no-marketing`, email verification, anti-bot) tracked in PRD-530.
- Kept in JS to mirror `login.js`; can move to TS if preferred.
- No non-interactive/CI fail-fast mode yet (prompts on missing fields, like `login`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add `forest signup` CLI command to create a Forest Admin account
> - Adds a new `signup` oclif command in [`src/commands/signup.js`](https://github.com/ForestAdmin/toolbelt/pull/768/files#diff-9745da2aa3abd54c60d5071b2bc47450fdbc3e0259925b39d1ba5a754a1e44e6) with optional `--email`, `--password`, `--first-name`, and `--last-name` flags; missing values are collected interactively via prompts.
> - Adds `Api.signup` in [`src/services/api.js`](https://github.com/ForestAdmin/toolbelt/pull/768/files#diff-64e6a6dc6ddee0b3b28d21c6b2bf6a5b1933a1a0a0f3284c9a3df46fea8bbf2e) that POSTs a JSON:API payload to `/api/users` to create the account, then logs in immediately via `/api/sessions` and persists the session token.
> - Error mapping in `Authenticator.toSignupError` surfaces server-provided messages for known failures (e.g. 409 email taken, 422 weak password) and a standard rate-limit message on 429.
>
> 
>
[...]

PR #770, opened 2026-06-14
feat(users): add forest users:invite command

Gawtier:
## What

Adds `forest users:invite` to invite users to a project from the CLI (`POST /api/invitations`).

```
forest users:invite -e <email> --team <name> --role <name> -l <permission-level> [-p <projectId>]
```

- `--email/-e` is repeatable (invite several users at once).
- `--permission-level/-l` ∈ `admin|editor|user|developer|manager`.

## Self-sufficient by design

No opaque IDs: **team and role are resolved by name** — auto-picked when a single one exists, prompted when several, with a clear error listing valid names when unknown. The project is resolved the usual way (`-p`, `FOREST_ENV_SECRET`, or prompt). So it works headlessly (names in flags) or interactively (pick from lists).

Resolution goes through two **new read-only managers** — `team-manager` / `role-manager` (`listForProject` → `GET /api/projects/:projectId/teams` and `/roles`). Future `teams:*` / `roles:*` (PRD-528/529) should **extend these** rather than re-implement the HTTP calls (coordination notes left on those issues).

## Validated end-to-end

Exercised on real prod as part of the full headless flow: signup → create project → boot dev agent → create prod env → deploy to Heroku → `users:invite` → invitation delivered. Note: a project's first role is created on production deployment, so invite requires the prod env to be deployed first (or a role created via `POST /api/roles`).

## Tests

- Command tests: success (name resolution), unknown team name (lists available), server error relayed.
- `yarn build` OK · full lint 0 errors · suites green.

Refs PRD-525.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add `users:invite` CLI command to invite users to a Forest project
> - Adds a new `users:invite` command in [`src/commands/users/invite.ts`](https://github.com/ForestAdmin/toolbelt/pull/770/files#diff-34d2442af9c1bf123dfd48ab0fd1d80a489770d114f82a536d7556ea9ddf6586) that accepts one or more `--email` addresses, a `--permission-level`, and optional `--team` and `--role` flags.
> - When `--team` or `--role` are omitted or ambiguous, the command fetches available teams/roles and prompts interactively; if a provided name is not found, it exits with an error listing valid options.
[...]

PR #772, opened 2026-06-15
feat(signup): add forest signup command

Gawtier:
Adds `forest signup` — account creation (`POST /api/users` + session), token persisted like `forest login`. Server errors relayed verbatim (generic on an existing email for anti-enumeration). Command + unit tests; README updated. Validated e2e. Linear **PRD-523**.

(Intra-repo re-open of fork PR #768 — CI now has secrets.)

> [!NOTE]
> ### Add `forest signup` CLI command to create a Forest Admin account
> - Adds a new `signup` oclif command in [`src/commands/signup.js`](https://github.com/ForestAdmin/toolbelt/pull/772/files#diff-9745da2aa3abd54c60d5071b2bc47450fdbc3e0259925b39d1ba5a754a1e44e6) with optional `--email`, `--password`, `--first-name`, and `--last-name` flags; missing values are collected interactively via prompts.
> - Adds `Api.signup` in [`src/services/api.js`](https://github.com/ForestAdmin/toolbelt/pull/772/files#diff-64e6a6dc6ddee0b3b28d21c6b2bf6a5b1933a1a0a0f3284c9a3df46fea8bbf2e) that POSTs a JSON:API payload to `/api/users` to create the account, then logs in immediately via `/api/sessions` and persists the session token.
> - Error mapping in `Authenticator.toSignupError` surfaces server-provided messages for known failures (e.g. 409 email taken, 422 weak password) and a standard rate-limit message on 429.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 507eb8b.</sup>
> 
>

### DIFF
--- a/src/commands/signup.js
+++ b/src/commands/signup.js
@@ -1,0 +1,42 @@
+const { Flags } = require('@oclif/core');
+const AbstractCommand = require('../abstract-command').default;
+
+class SignupCommand extends AbstractCommand {
+  constructor(argv, config, plan) {
+    super(argv, config, plan);
+    const { assertPresent, authenticator } = this.context;
+    assertPresent({ authenticator });
+    this.authenticator = authenticator;
+  }
+
+  async run() {
+    const { flags } = await this.parse(SignupCommand);
+    await this.authenticator.trySignup({
+      email: flags.email,
+      password: flags.password,
+      firstName: flags['first-name'],
+      lastName: flags['last-name'],
+    });
+  }
+}
+
+SignupCommand.description = 'Create a new Forest Admin account.';
+
+SignupCommand.flags = {
+  email: Flags.string({
+    char: 'e',
+    description: 'Your Forest Admin account email.',
+  }),
+  password: Flags.string({
+    char: 'P',
+    description: 'Your Forest Admin account password.',
+  }),
+  'first-name': Flags.string({
+    description: 'Your first name.',
+  }),
+  'last-name': Flags.string({
+    description: 'Your last name.',
+  }),
+};
+
+module.exports = SignupCommand;
--- a/src/services/authenticator.js
+++ b/src/services/authenticator.js
@@ -122,6 +122,91 @@ function Authenticator({
    * }} params
    * @returns {Promise<string>}
    */
+  this.trySignup = async config => {
+    await this.logout({ log: false });
+    try {
+      const token = await this.signup(config);
+      await this.saveToken(token);
+      logger.info('Account created. You are now logged in.');
+    } catch (error) {
+      const message =
+        error instanceof ApplicationError
+          ? error.message
+          : `${ERROR_UNEXPECTED} ${chalk.red(error)}`;
+      logger.error(message);
+    }
+  };
+
+  /**
+   * Create a new account, then open a session and return its token.
+   * @param {{
+   *  email: string;
+   *  password: string;
+   *  firstName: string;
+   *  lastName: string;
+   * }} params
+   * @returns {Promise<string>}
+   */
+  this.signup = async ({ email, password, firstName, lastName }) => {
+    if (email) {
+      const validationResult = await this.validateEmail(email);
+      if (validationResult !== true) {
+        throw new ApplicationError(validationResult);
+      }
+    } else email = await this.promptEmail();
+
+    if (!firstName) firstName = await this.promptRequired('firstName', 'What is your first name?');
+    if (!lastName) lastName = await this.promptRequired('lastName', 'What is your last name?');
+    if (!password) {
+      ({ password } = await inquirer.prompt([
+        {
+          type: 'password',
+          name: 'password',
+          message: 'Choose a password:',
+          validate: input => !!input || 'Please enter a password.',
+        },
+      ]));
+    }
+
+    try {
+      await api.signup({ email, password, firstName, lastName });
+    } catch (error) {
+      throw this.toSignupError(error);
+    }
+
+    return api.login(email, password);
+  };
+
+  this.promptRequired = async (name, message) => {
+    const answer = await inquirer.prompt([
+      {
+        type: 'input',
+        name,
+        message,
+        validate: input => !!input || 'This field is required.',
+      },
+    ]);
+    return answer[name];
+  };
+
+  this.toSignupError = error => {
+    const status = error.status || (error.response && error.response.status);
+    const detail =
+      error.response && error.response.body && error.response.body.errors
+        ? error.response.body.errors[0] && error.response.body.errors[0].detail
+        : null;
+
+    // Relay the server message verbatim: it intentionally returns a generic
+    // message when the email already exists (anti user-enumeration) and a
+    // specific one for validation errors (e.g. weak password). We surface
+    // whatever it chose to say rather than inferring from the status code.
+    if (detail) return new ApplicationError(detail);
+    if (status === 429) {
+      return new ApplicationError('Too many attempts. Please wait a moment and try again.');
+    }
+    return error;
+  };
+
   this.login = async ({ email, password, token }) => {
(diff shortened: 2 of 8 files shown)


src/services/authenticator.js
```diff
@@ -122,6 +122,91 @@ function Authenticator({
...
+      await this.saveToken(token);
+      logger.info('Account created. You are now logged in.');
+    } catch (error) {
+      const message =
+        error instanceof ApplicationError
+          ? error.message
+          : `${ERROR_UNEXPECTED} ${chalk.red(error)}`;
+      logger.error(message);
+    }
+  };
```
matthv, line 138: This catch logs the error but never rethrows or sets a non-zero exit code, so forest signup returns 0 even when the account wasn't created. The "email already taken (409)" test confirms it: it asserts the stderr message but no exitCode.

It's fine for interactive login, but signup is a primitive of the headless orchestrator (PRD-526, "skills decide, the CLI executes"), which chains commands and relies on exit codes. A failed signup that returns 0 makes the orchestrator think the account exists — the failure only resurfaces later at projects:create as a confusing auth error.

Note it also diverges from users:invite (#770), which already does this.exit(1) on error — the two onboarding primitives have opposite exit conventions today.

Suggested fix: rethrow from trySignup (or this.exit(1) in the command's catch) so failures exit non-zero, and add exitCode: 1 to the failure tests signupEmailTaken, signupWeakPassword, invalid email). Worth deciding whether to align login too, or keep the divergence intentionally (interactive vs orchestrated).

src/commands/signup.js
```diff
@@ -0,0 +1,42 @@
...
+
+SignupCommand.flags = {
+  email: Flags.string({
+    char: 'e',
+    description: 'Your Forest Admin account email.',
+  }),
+  password: Flags.string({
+    char: 'P',
+    description: 'Your Forest Admin account password.',
+  }),
```
matthv, line 33: A password passed as -P 'secret' leaks into shell history, the process table (ps aux), and CI logs. For a security-sensitive product and especially for  a primitive driven by the headless orchestrator.  I'd avoid argv as a secret channel.
The ecosystem consensus: keep the masked interactive prompt as the default (already done here ✅), and offer a non-interactive but safe channel via stdin docker login --password-stdin, gh auth login --with-token both do this). Stdin is preferable to an env var here because the orchestrator later spawnsnpm/heroku, and an env var like FOREST_PASSWORD would be inherited by those child processes — stdin is consumed and gone.
Proposal — add `--password-stdin`:
Flag (mutually exclusive with --password):
```
 'password-stdin': Flags.boolean({
   description: 'Read the password from stdin (safer for scripts/CI than --password).',
   exclusive: ['password'],
 }),
```

  A small stdin reader:
```
  async function readPasswordFromStdin() {
    if (process.stdin.isTTY) {
      throw new ApplicationError('--password-stdin was set but nothing was piped to stdin.');
    }
    const chunks = [];
    for await (const chunk of process.stdin) chunks.push(chunk);
    // strip a single trailing newline (added by `echo`) — not a full trim,
    // a password may legitimately end with a space
    return Buffer.concat(chunks).toString('utf8').replace(/\r?\n$/, '');
  }
```

Wiring in run():
```
const password = flags['password-stdin'] ? await readPasswordFromStdin() : flags.password;
  await this.authenticator.trySignup({
    email: flags.email,
    password,
    firstName: flags['first-name'],
    lastName: flags['last-name'],
  });
```
  
  Usage:
```
  echo "$PWD" | forest signup -e me@mail.com --first-name M --last-name T --password-stdin
  forest signup -e me@mail.com --password-stdin < password.txt
```
Decision needed on -P: either drop it, or keep it for convenience but emit an insecurity warning the way docker does ("--password via flag is insecure, prefer --password-stdin"). My preference: keep the prompt as default, add --password-stdin, and at least warn on -P.
(Edge cases handled above: TTY-with-nothing-piped → clear error instead of hanging; single trailing newline stripped, not a global trim; exclusive prevents -P + --password-stdin together.)

src/services/authenticator.js
```diff
@@ -122,6 +122,91 @@ function Authenticator({
...
+    // Relay the server message verbatim: it intentionally returns a generic
+    // message when the email already exists (anti user-enumeration) and a
+    // specific one for validation errors (e.g. weak password). We surface
+    // whatever it chose to say rather than inferring from the status code.
+    if (detail) return new ApplicationError(detail);
+    if (status === 429) {
+      return new ApplicationError('Too many attempts. Please wait a moment and try again.');
+    }
+    return error;
+  };
```
matthv, line 208: if (detail) return new ApplicationError(detail) fires for any status. That's exactly right for 4xx (the anti-enumeration message on 409, validation on 422 — both meant for the user). But on a 5xx, a server-provided detail (internal message/trace) would be surfaced verbatim to the user instead of falling through to the generic "unexpected error" path.

 Suggestion: gate the detail relay to 4xx so 5xx stays generic:
 if (detail && status >= 400 && status < 500) return new ApplicationError(detail);
 Low priority / defensive — depends on what the backend actually returns on 5xx.

matthv (commented):
Nice work on this

Overall I think it's mergeable on substance — I have one point I'd consider blocking for the headless use case, plus a few suggestions, all left inline:
 - 🔴 Exit code — signup exits 0 even on failure, which breaks the orchestrator's ability to detect a failed step (and diverges from users:invite). Thisis the one I'd want resolved before merge.
 - 🟠 Password via -P flag — secret leaks into shell history / ps / CI logs; proposing a --password-stdin channel (with code).
 - 🟡 toSignupError on 5xx — relays the server detail regardless of status; suggest gating it to 4xx.
- 🟡 Grouped minor remarks — unused api.signup return value, hand-rolled JSON:API payload, whitespace-only name validation.





A few small things, none blocking:
1. api.signup returns a value that's never used. It resolves to response.body (the created user), but Authenticator.signup does await api.signup(...) and ignores the result before calling api.login. Either drop the return from api.signup, or use it (e.g. to skip the extra /api/sessions round-trip if the create response already carries what we need).
 2. JSON:API payload is hand-rolled in api.signup. Every other method in api.js goes through a serializer (environmentSerializer,  applicationTokenSerializer, …), whereas signup builds { data: { type: 'users', attributes: { … } } } inline. Fine if there's no user serializer, but it's a small departure from the file's convention — worth a serializer for consistency if one is cheap to add.
3. Name validation accepts whitespace-only input. In promptRequired, validate: input => !!input || '…' lets a string of spaces through (it's truthy).  Same pre-existing pattern as login's password prompt, so it's consistent — but a .trim() would make first/last name validation actually meaningful:  validate: input => !!input.trim() || 'This field is required.',
For reference, the JS-vs-TS choice for signup.js is already covered in the PR description, so not repeating it here.)

Gawtier:
Closing this for now — after the review (thanks @matthv, your feedback is exactly what led here) we've decided **not to ship signup in the CLI**. Account creation stays in the UI; the CLI's auth story stops at `forest login`.

Reasoning:
1. **Consent capture (T&Cs / Privacy Policy).** Account creation requires legally-valid acceptance of the terms — proof that *this* user accepted *this* version, on *this* date. The UI does this natively with an audit trail; replicating it in a headless CLI flow (an agent "accepting" on a user's behalf) is both extra work and legally shaky.
2. **Pwd in clear in CLI**. No obvious way to tackle this (as the stdin solution won't work with LLMs)
3. **Not on the critical path.** Signup is a one-shot, per-user action. The recurring primitive the headless orchestrator actually needs is `login`; the orchestrator's value starts *after* auth. We'll bootstrap the user through UI signup once, then let the CLI take over.

Next steps (separate, small PR): the onboarding skill will detect the unauthenticated state and point users to sign up at app.forestadmin.com before `forest login`, and we may enrich the `forest login` failure message with a "no account yet? sign up at <url>" hint. No signup code, no secret handling, no legal surface.

PR #773, opened 2026-06-15
feat(environments): support --type and optional URL on environments:create

Gawtier:
Exposes `--type production` and makes `--url` optional; payload sends `apiEndpoint`/`type` only when provided. Aligned with the server (apiEndpoint nullable for all types, prod-creation idempotent, ordering enforced server-side). Linear **PRD-524**.

(Intra-repo re-open of fork PR #769.)

> [!NOTE]
> ### Support `--type` flag and optional `--url` on `environments:create` command
> - Makes `--url`/`-u` optional in `environments:create`, allowing environments to be created without an API endpoint (e.g. production environments).
> - Adds `--type`/`-t` flag accepting `'production'` or `'remote'` values.
> - `EnvironmentManager.createEnvironment` now conditionally includes `apiEndpoint` and `type` in the POST payload based on what flags are provided.
> - Behavioral Change: the POST to `/api/environments` no longer always sends `apiEndpoint`; omitting `--url` will result in a payload without that field.
>
> 
> #### Changes since #773 opened
>
> - Added validation in `CreateCommand.runAuthenticated` method to enforce URL requirements based on environment type and prevent incompatible flag combinations [dc7cdc2]
> - Added test coverage for environment creation validation logic including production environments with URLs, remote environments without URLs, and production environments with role disablement [dc7cdc2]
> - Added comment in `EnvironmentManager` function clarifying server default behavior for environment type [dc7cdc2]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c8ead6e.</sup>
> 
>
>

### DIFF
--- a/src/commands/environments/create.js
+++ b/src/commands/environments/create.js
@@ -15,6 +15,23 @@ class CreateCommand extends AbstractAuthenticatedCommand {
   async runAuthenticated() {
     const parsed = await this.parse(CreateCommand);
     const config = await withCurrentProject({ ...this.env, ...parsed.flags });
+
+    const isProduction = config.type === 'production';
+
+    // Only production environments may be created without a URL; a remote env
+    // without an apiEndpoint would be created permanently inactive.
+    if (!isProduction && !config.url) {
+      this.logger.error('The --url option is required unless --type production is set.');
+      this.exit(1);
+    }
+
+    // Roles cannot be disabled on production: the server skips role creation entirely
+    // when areRolesDisabled is true, which would leave the project without its default roles.
+    if (isProduction && config.disableRoles) {
+      this.logger.error('The --disableRoles option cannot be used with --type production.');
+      this.exit(1);
+    }
+
     const manager = new EnvironmentManager(config);
 
     try {
@@ -55,7 +72,7 @@ CreateCommand.flags = {
   url: Flags.string({
     char: 'u',
     description:
-      'Application URL. Optional: the environment is created inactive until an apiEndpoint is set.',
+      'Application URL. Required, except with --type production where it may be omitted.',
   }),
   type: Flags.string({
     char: 't',
--- a/src/commands/environments/create.js
+++ b/src/commands/environments/create.js
@@ -54,8 +54,12 @@ CreateCommand.flags = {
   }),
   url: Flags.string({
     char: 'u',
-    description: 'Application URL.',
-    required: true,
+    description: 'Application URL (optional for a production environment).',
+  }),
+  type: Flags.string({
+    char: 't',
+    description: 'Environment type. Use "production" to create the first production environment.',
+    options: ['production', 'remote'],
   }),
   format: Flags.string({
     char: 'format',
--- a/src/commands/environments/create.js
+++ b/src/commands/environments/create.js
@@ -54,11 +54,13 @@ CreateCommand.flags = {
   }),
   url: Flags.string({
     char: 'u',
-    description: 'Application URL (optional for a production environment).',
+    description:
+      'Application URL. Optional: the environment is created inactive until an apiEndpoint is set.',
   }),
   type: Flags.string({
     char: 't',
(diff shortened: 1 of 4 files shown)


matthv (commented):
Solid PR — cleaner than the signup one, and nicely done: errors exit with a non-zero code (great for the headless flow), and it reuses the existing serializer. No blocker here. A few things worth tightening, mostly around the new option combinations and tests. I've explained each one so it's easy to ct on 👇

  ---
  1. --type production combined with --disableRoles looks inconsistent

  What: --disableRoles is always forwarded as areRolesDisabled, including for a production environment. In the existing code, areRolesDisabled is only ever
  treated as meaningful for remote environments (see renderers/environments.js), which suggests disabling roles isn't a production concept.

  Why it matters: A project's first role ("Operations") is created when production is deployed, and users:invite depends on that role. Creating a production
  env with roles disabled is at best meaningless and could interfere with that — though I haven't confirmed the exact server behavior.

  Suggestion: Could you check what the server does with areRolesDisabled: true on a production env? If it's not a valid combination, ignoring --disableRoles
  when type === 'production' (or rejecting it with a clear message) would prevent a confusing footgun.

  2. -u (URL) is now optional for every environment type, not just production

  What: The ticket asked to make the URL optional for production. The change makes it optional for remote environments too (it used to be required).
  Why it matters: You can now create a remote environment with no URL — which leaves it permanently inactive until someone runs environments:update. That may
  be fine, but it's a behavior change beyond production, and we've only verified the production case end-to-end.
  Suggestion: Confirm the server actually accepts a remote env with no URL, and decide if that's intended. If only production should allow a missing URL, we
  could keep requiring it for remote.

  3. The default type ("remote") is decided by the server, not the CLI

  What: When --type is omitted, the CLI sends nothing and the server falls back to "remote". The flag description says "defaults to remote".
  Why it matters: It's an implicit contract — if the server's default ever changes, the CLI's help text becomes wrong without anyone touching the CLI. Not a
  bug, just something to be aware of.
  Suggestion: Either leave a short code comment noting the default comes from the server, or set type: 'remote' explicitly in the CLI so the behavior is
  self-contained.

  4. Add a couple of tests for the new option combinations

  What: There's one new test (--type production with no URL). The change opens up cases that aren't covered yet.
  Why it matters: These are exactly the new behaviors this PR introduces, so they're the ones most worth locking in.
  Suggestion: Add tests for: (a) production with a URL (does apiEndpoint get sent alongside type?), and (b) remote with no URL (the newly-allowed case). The
  "invalid type is rejected" case is already handled by oclif, so no need to test that.

Gawtier:
Thanks @matthv — all four addressed in dc7cdc2. I checked the server (`private-api`) to settle the open questions instead of guessing:

**1. `--type production` + `--disableRoles` — confirmed footgun.** In `environment-service.ts`, `areRolesDisabled` is honored for *every* type, production included: `createEnvironmentInternal` skips `createPermissionsForEnvironment` entirely when it's `true`, so no default roles get created. The server doesn't reject it, it silently obeys. → The CLI now **rejects** `--disableRoles` with `--type production` (exit 1, clear message), consistent with the rest of the headless flow.

**2. `-u` optional beyond production — server accepts it, but we're restricting anyway.** `checkCanCreateRemoteEnvironment` only validates plan limits, not URL presence, and `apiEndpoint` is a nullable column (URL validators only run when a value is present). So a remote with no URL *is* accepted server-side — created permanently inactive, exactly as you described. Since PRD-524 only asked for production, I kept it conservative: **URL is now required unless `--type production`**, validated CLI-side. Flag description updated to match.

**3. Default `remote` comes from the server — agreed.** Added a short comment in `environment-manager` noting the default originates server-side rather than duplicating the contract in the CLI.

**4. Tests added:** production *with* a URL (asserts `apiEndpoint` is sent alongside `type`), remote *without* a URL (now rejected), and production + `--disableRoles` (rejected). 10/10 green.

matthv (approved):

PR #774, opened 2026-06-15
feat(users): add forest users:invite command

Gawtier:
Adds `forest users:invite` (`POST /api/invitations`). Team & role are resolved **by name** (auto-pick / prompt / clear error) via new read-only `team-manager`/`role-manager` so PRD-528/529 can extend them. Validated e2e. Linear **PRD-525**.

(Intra-repo re-open of fork PR #770.)

> [!NOTE]
> ### Add `users:invite` CLI command to batch-invite users to a project
> - Adds a new `users:invite` command in [invite.ts](https://github.com/ForestAdmin/toolbelt/pull/774/files#diff-34d2442af9c1bf123dfd48ab0fd1d80a489770d114f82a536d7556ea9ddf6586) that accepts one or more `--email` flags along with `--team`, `--role`, and `--permission-level` options.
> - Resolves team and role by name: auto-selects when only one exists, prompts interactively when multiple exist, and exits with an error listing valid names when a provided name is not found.
> - Adds three supporting services: [InvitationManager](https://github.com/ForestAdmin/toolbelt/pull/774/files#diff-3f1db832a0d5b6850f1f157c140706742b2d5e1153f802934469d10bf47b5d19) (POST `/api/invitations`), [TeamManager](https://github.com/ForestAdmin/toolbelt/pull/774/files#diff-7083afebe926c101165a5a4bd58e5f3528e9eca5990974175bd3f4dbb6da1019) (GET project teams), and [RoleManager](https://github.com/ForestAdmin/toolbelt/pull/774/files#diff-9a0caf92ffc6f1d4735786395cc30a13986ca93e439a2e374f466da135171ad0) (GET project roles).
> - Supports `--format json` for machine-readable output and surfaces server-provided error details (e.g. 422 responses) with exit code 1.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f412cec.</sup>
> 
>

### DIFF
--- a/src/commands/users/invite.ts
+++ b/src/commands/users/invite.ts
@@ -1,0 +1,143 @@
+import type { Config } from '@oclif/core';
+
+import { Flags } from '@oclif/core';
+
+import AbstractAuthenticatedCommand from '../../abstract-authenticated-command';
+import InvitationManager from '../../services/invitation-manager';
+import RoleManager from '../../services/role-manager';
+import TeamManager from '../../services/team-manager';
+import withCurrentProject from '../../services/with-current-project';
+
+const PERMISSION_LEVELS = ['admin', 'editor', 'user', 'developer', 'manager'];
+
+type NamedEntity = { id: number | string; name: string };
+
+export default class UsersInviteCommand extends AbstractAuthenticatedCommand {
+  private env: Record<string, string>;
+
+  private inquirer: { prompt: (questions: unknown[]) => Promise<Record<string, unknown>> };
+
+  static override description = 'Invite users to a project.';
+
+  static override flags = {
+    email: Flags.string({
+      char: 'e',
+      description: 'Email of the user to invite (repeat the flag to invite several users).',
+      required: true,
+      multiple: true,
+    }),
+    team: Flags.string({
+      char: 't',
+      description: 'Team name to add the invited users to (prompted if omitted and several exist).',
+    }),
+    role: Flags.string({
+      char: 'r',
+      description: 'Role name to assign (prompted if omitted and several exist).',
+    }),
+    'permission-level': Flags.string({
+      char: 'l',
+      description: 'Permission level of the invited users.',
+      options: PERMISSION_LEVELS,
+      required: true,
+    }),
+    projectId: Flags.integer({
+      char: 'p',
+      description: 'Forest project ID.',
+    }),
+    format: Flags.string({
+      char: 'f',
+      description: 'Output format.',
+      options: ['table', 'json'],
+      default: 'table',
+    }),
+  };
+
+  constructor(argv: string[], config: Config, plan?) {
+    super(argv, config, plan);
+    const { assertPresent, env, inquirer } = this.context;
+    assertPresent({ env, inquirer });
+    this.env = env;
+    this.inquirer = inquirer;
+  }
+
+  // Resolve a team/role by name (headless when passed as a flag), with the
+  // same UX as project selection: auto-pick when there is only one, prompt
+  // when several and none provided, clear error listing the valid names.
+  private async resolveByName(
+    entities: NamedEntity[],
+    providedName: string | undefined,
+    label: string,
+  ): Promise<NamedEntity> {
+    if (providedName) {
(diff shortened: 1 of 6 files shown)


matthv:
Really nice PR — this is the most polished of the three. Exit codes are correct (failures exit non-zero, and 403s are deliberately handed off to the base command's handler — good call), errors are relayed cleanly, and the "resolve team/role by name" logic is well done (auto-picks when there's only one, lists the valid names on a typo). **No blocker.** A few things worth tightening, explained below 👇

---

**1. The "no role yet" case works — let's make its message more actionable (and test it)**

*What:* When a project has no role yet (i.e. production hasn't been deployed), inviting fails cleanly with `No role found in this project.` and a non-zero exit code. That's exactly the right behavior — it's the ordering gate from the test plan (you can't invite before the first role exists, which is created on production deploy).

*Why it matters:* For someone driving this headlessly, the message says *what's* wrong but not *what to do next*. A one-line hint turns a dead end into a next step.

*Suggestion:* When the missing entity is specifically the role, point them forward — e.g. `No role found in this project. Deploy a production environment first (it creates the project's first role), or create one via the API.` And since this is a documented gate, it deserves a test (a roles-list that comes back empty → expect the message + exit 1).

**2. Mixed TypeScript / JavaScript within the same PR**

*What:* The command itself is TypeScript (`invite.ts`), but the three new services it introduces (`invitation-manager.js`, `role-manager.js`, `team-manager.js`) are plain JavaScript.

*Why it matters:* JS services are consistent with the existing `environment-manager.js`, so it's not wrong — but shipping a new TS command alongside new JS services in one PR is a bit inconsistent given the codebase is moving toward TS. It's the kind of thing that's cheap to align now and annoying to migrate later.

*Suggestion:* Either write the three managers in TypeScript too (preferred, to match the direction of travel), or consciously decide "services stay JS for now" and keep it consistent. Whichever — just an intentional choice rather than a mix by accident.

**3. A few tests would lock in the behaviors this PR adds**

*What:* There are three solid tests (happy path, unknown team name, server rejection). A few key paths aren't covered yet.

*Why it matters:* These are exactly the behaviors that make this command useful headlessly, so they're worth pinning down so a future change can't silently break them.

*Suggestion:* Add tests for:
- **multiple invitations** — `-e` passed several times (the `multiple: true` path), to confirm all emails end up in the request;
- **the "no role" gate** from point 1 (empty roles list → message + exit 1);
- **`--format json`** — that it prints the raw result.

---

Minor (non-blocking): in `resolveByName`, after `this.exit(1)` there's no `return`. It works today because `exit` throws (it's typed `never`), but a defensive `return` after each `this.exit(1)` would make the control flow obvious and robust if `exit` ever gets stubbed in a test.

Gawtier:
Thanks for the thorough review 🙏 Pushed f412cec addressing it:

**1. "No role yet" → actionable + tested.** ✅ The empty-role case now points forward: `No role found in this project. Deploy a production environment first (it creates the project's first role), or create one via the API.` Implemented as an optional `emptyHint` on `resolveByName` (kept generic for `team`, hint only for `role`), and covered by a new test (empty roles list → message + exit 1).

**2. Mixed TS/JS.** Deliberate choice for this PR: the three managers stay JS to match the existing `environment-manager.js`, and I'd rather not mix a feature with a TS migration in the same diff. Happy to migrate the services in a dedicated follow-up so it's a clean, reviewable change rather than noise here.

**3. More tests.** ✅ Added: multiple invitations (`-e` repeated → both emails in one request), the no-role gate (point 1), and `--format json` (raw result printed).

**4. `return` after `this.exit(1)`.** ✅ Added in `resolveByName` (control flow is now explicit even if `exit` is stubbed). Note: I left the `this.exit(1)` in `runAuthenticated`'s catch as-is — adding a `return` there trips `consistent-return` because the JSON branch already does a bare `return;`.

**Bonus (not flagged):** hardened the catch — `JSON.parse(response.text)` is now guarded, so a non-JSON error body rethrows the original error instead of surfacing a cryptic `SyntaxError`.

Lint + the full `users:invite` suite (6 tests) are green.

matthv (approved):
LGTM

PR #776, opened 2026-06-16
feat(layout): add forest layout pull/diff/apply commands

Gawtier:
## What

Declarative **layout-as-code** for the CLI under `forest layout` — **create *and* edit** layout elements in one pull → diff → apply flow:

- **`forest layout pull --env --team [--out]`** — read the live rendering + folders + workflows into a versionable `forest-layout.yml`.
- **`forest layout diff [file]`** — show the JSON-Patch plan the file implies (dry-run, reads only).
- **`forest layout apply [file]`** — push the changes as JSON-Patch arrays to `PATCH /api/{layout,folders,workflows}`. Idempotent: re-applying an unchanged file is a no-op.

Create or edit **collections** (scalars, fields, columns), **segments**, **dashboards**, **charts**, **workspaces** (+ components), **inboxes**, and **workflows** — including a workflow's **internal step graph** (compiled to BPMN). Scope resolves from `--env`/`--team` (id or name) → file header → defaults, reusing the toolbelt's auth/managers (no new auth).

## How

- **Pure engine (no IO, unit-tested):** rendering→canonical mapper; identity diff (match by `id` then `name`, `add→replace→remove`); opaque pass-through for complex values; a mirror of the server patch whitelist that pre-validates every op.
- **Idempotent & partial-authoring friendly:** a field absent from the file means "leave alone" (set `null` to clear), so hand-authored files needn't carry every field. The create path keeps the client-supplied `id` (generates a uuid when absent) and fills the write-required fields the read omits.
- **Workflow logic:** a workflow's `steps` graph is compiled to BPMN and uploaded exactly as the Forest UI does (verified against a request trace): `generate-presigned-request` → multipart S3 upload → link via `bpmnAwsS3Identifier`. Before uploading, the compiled BPMN is byte-compared against the stored version and **skipped if unchanged** — re-apply stays a no-op.
- **Multi-domain:** one round covers `layout` (incl. workspaces/inboxes) + `folders` + `workflows`; apply sends one atomic PATCH per domain.

## Validation

- 68 unit tests (`test/services/layout/`).
- **Live, end to end on a real env:** created a segment, dashboard, workspace, inbox, a workflow shell, and a **workflow with a step graph** (compiled → BPMN uploaded → linked); edited scalars; **re-applied a workflow-with-steps file twice → second run is a clean no-op** (no BPMN re-upload). Pull round-trips 12 workspaces (250 components) + 6 inboxes + 12 workflows idempotently. Every change reverted cleanly.

## Notes

- Idempotent management relies on stable ids: pulled files always carry them; hand-authored *new* items should include an `id` to be managed across applies (otherwise each apply creates a fresh one).
- name↔id resolution inside complex values isn't implemented — and is largely moot here since ids are client-supplied and kept stable, so embedded references stay portable across envs.

## Dependency

⚠️ **Depends on #774** (`team-manager.js`, reused by the scope resolver). Lint/Build CI is red until #774 merges into `main`; this PR then goes green without a rebase.

## Linear

Implements **PRD-555** (engine, HTTP adapter, scope resolver, commands, tests). Absorbs the create-path value work (ex-PRD-560) and the workflow authoring (ex-PRD-561).


matthv:
## Code review — `forest layout pull / diff / apply` (PRD-555 + absorbs ex-560/561)

Excellent, professional-grade work: clean pure-engine → IO-adapter → command separation, careful idempotency (deterministic BPMN compiler + byte-compare skip), proper XML escaping, actionable 422/403 messages, headless defaults that fail loudly instead of prompting, and a defensive whitelist self-check. The pure engine is well tested (68 tests).

**The PR works (validated e2e).** Two items worth doing before merge (**D1**, **T1**); **P2** is a merge-order gate (no code change).

### 🟡 Should-fix

**D1 — Name-only edits are silently destructive.** `diff.ts` says items match *"by `id`, falling back to `name`"*, but `indexByIdentity` keys them `id:<x>` or `name:<x>` — separate namespaces. Pulled items always have an id, so a hand-authored item carrying only `name` matches nothing and is treated as **add (new uuid) + remove (old)**, discarding server state bound to the old id (segment columns/filter, dashboard chart ids, a workflow's BPMN…) or creating a duplicate. The only guard today is a doc comment. Fix: index remote under both `id:` and `name:` keys (or fail loudly on a name collision without id), and correct the header comment. Critical for **PRD-559**, which will *generate* these files.

**T1 — No tests for the IO layer or orchestration.** The 68 tests cover the pure engine; untested are the trickiest parts: `sync.planWorkflowBpmn` (the BPMN changed/unchanged/error decision), `apply` orchestration (ids mutated before diff, per-domain ordering, BPMN-only path), and `layout-manager` (presigned-S3 upload, error mapping). A `planWorkflowBpmn` test + one `apply` idempotency test would lock what's most likely to regress.

### 🟢 Style — comment density (optional)

The engine is **heavily** commented — most functions carry multi-line JSDoc plus inline narration. A lot of it is genuinely valuable *"why"* and should stay (e.g. the partial-authoring semantics in `diff.ts:117-120`, the BPMN byte-compare idempotency in `sync.ts`, the "mirrors the Forest UI" notes on the S3 upload in `layout-manager.ts`). The trim target is the **redundant / duplicated** prose:

- **Same rationale repeated across files.** The "there is no raw GET for the layout domain — `GET /api/layout` returns `[]`" explanation appears at least three times: `layout-manager.ts` class header (l. 8-22), `layout-manager.ts` `getLayoutDomain` doc (l. 61-65), and `rendering-mapper.ts` header (l. 1-13). State it once (one canonical spot) and let the others reference it in a line.
- **JSDoc that restates the signature.** Several method docs re-describe what the name + types already say (`GET /api/… — the rendering as JSON:API`, etc.). Keep the one-liner where it adds the route/headers; drop the parts that just paraphrase the code.
- **Rule of thumb for the managed agent:** keep comments that explain a *non-obvious decision or server constraint*; cut comments that explain *what the next line does*. Net effect is higher signal, not less documentation.

Low priority — purely readability, no behaviour involved.

### Process
- **P2 — Depends on #774 (merge-order gate).** CI is red until #774 (`team-manager.js`) lands on `main`. Merge it first, re-confirm green.
- **P1 — Scope (awareness).** Goes well beyond PRD-555's "simple element types" — absorbs ex-560 (opaque values) and ex-561 (workflows→BPMN→S3, inboxes, workspaces, charts), +3188 LOC. QA should scope to the full surface.

### Naming / structure
Excellent and consistent. 👍

matthv:
Moved to the upstream repo to take over the feature: #777. It carries the full history of this branch plus the review fixes (D1 identity matching, T1 orchestration tests). Closing in favour of #777 — the review thread here still applies.

PR #795, opened 2026-07-02
feat(layout): add `forest layout patch` command

Gawtier:
## What
Adds `forest layout patch` — apply a JSON-Patch directly to a layout (one PATCH per domain, mirroring the front-end), instead of the `pull → edit → apply` round-trip.

## Why
For programmatic / agent-driven layout edits where the exact ops are already known, a direct patch is simpler and atomic per domain. Complements the existing `layout pull` / `layout apply`.

## How
- reads ops from a file arg or stdin (`-`)
- validates op type + path against the server whitelist (`patch-rules`)
- `--dry-run` prints the planned ops without sending anything
- scope via `-e/--env`, `-p/--projectId`, `-t/--team` (`resolveCommandScope`)
- `workflows` domain intentionally deferred to `layout apply` (creating a workflow needs a BPMN S3 upload, not a single layout PATCH)

## Tests
- 11 unit tests (`toPlannedOps` / `buildAllOps` / `formatOps`): op validation, whitelist rejection, per-domain ordering, workflows-deferred — all green.
- `eslint` + `tsc --noEmit` clean.

## Notes
- Command is auto-discovered by oclif (`src/commands/layout/patch.ts` → `forest layout patch`); `oclif.manifest.json` and the README command list are regenerated at pack time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add `forest layout patch` CLI command for applying JSON patches to layout domains
> - Adds a new authenticated `layout patch` command in [patch.ts](https://github.com/ForestAdmin/toolbelt/pull/795/files#diff-2957a72e606bfe75fd4b36f7a745fffa83bff16698a0912f488937fce57f11aa) that reads a JSON patch file (or stdin) and applies it per-domain via `LayoutManager.patchDomain`.
> - Validates input shape, op types, and path prefixes against a server-side whitelist before executing; supports `--dry-run` to preview planned operations without applying them.
> - Supports `layout` and `folders` domains (rejects `workflows`); executes domain patches sequentially and reports scoped errors with exit code 2 on non-401 API failures.
> - Unit tests cover `toPlannedOps`, `buildAllOps`, and `formatOps` edge cases including unknown ops, bad paths, whitelist rejection, and domain ordering.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0396a03.</sup>
> 
>

PR #796, opened 2026-07-02
feat(layout): add `forest layout patch` command

Gawtier:
## What
Adds `forest layout patch` — apply a JSON-Patch directly to a layout (one PATCH per domain, mirroring the front-end), instead of the `pull → edit → apply` round-trip.

## Why
For programmatic / agent-driven layout edits where the exact ops are already known, a direct patch is simpler and atomic per domain. Complements the existing `layout pull` / `layout apply`.

## How
- reads ops from a file arg or stdin (`-`)
- validates op type + path against the server whitelist (`patch-rules`)
- `--dry-run` prints the planned ops without sending anything
- scope via `-e/--env`, `-p/--projectId`, `-t/--team` (`resolveCommandScope`)
- `workflows` domain intentionally deferred to `layout apply` (creating a workflow needs a BPMN S3 upload, not a single layout PATCH)

## Tests
- 11 unit tests (`toPlannedOps` / `buildAllOps` / `formatOps`): op validation, whitelist rejection, per-domain ordering, workflows-deferred — all green.
- `eslint` + `tsc --noEmit` clean.

## Notes
- Command is auto-discovered by oclif (`src/commands/layout/patch.ts` → `forest layout patch`); `oclif.manifest.json` and the README command list are regenerated at pack time.
- Supersedes #795 (opened from a fork, which had no access to CI secrets → the DockerHub login step failed and blocked the Test job). This one is branched inside the org repo so CI runs with secrets.

> [!NOTE]
> ### Add `forest layout patch` command to apply JSON-patch operations to layout and folders
> - Adds a new `layout patch` CLI command in [patch.ts](https://github.com/ForestAdmin/toolbelt/pull/796/files#diff-2957a72e606bfe75fd4b36f7a745fffa83bff16698a0912f488937fce57f11aa) that reads patch input from a file or stdin, validates ops across the `layout` and `folders` domains, and applies one atomic PATCH per domain via `LayoutManager`.
> - Supports `--dry-run`, `--force`, `--env`, `--team`, and `--projectId` flags; non-interactive runs require all scope flags unless `FOREST_ENV_SECRET` is set.
> - Validates input as a non-null JSON object, checks each operation's shape, op type (`add`, `replace`, `remove`, `test`), and path against a whitelist before sending.
> - Prints a grouped, human-readable summary of planned ops before applying; prompts for confirmation unless `--force` is passed.
> - Risk: on a recoverable API error (`LayoutApiError` non-401), the command exits with code 2 and warns that already-applied domains may cause duplicate `add` replays if retried.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57bba44.</sup>
> 
>

### DIFF
--- a/src/commands/layout/patch.ts
+++ b/src/commands/layout/patch.ts
@@ -25,15 +25,37 @@ type PatchInput = Partial<Record<LayoutDomain, RawOp[]>>;
 
 const OP_PREFIX: Record<string, string> = { add: '+', remove: '-', replace: '~' };
 
+const STDIN_FIRST_BYTE_TIMEOUT_MS = 10_000;
+
 function readStdin(): Promise<string> {
   return new Promise((resolve, reject) => {
     let data = '';
+    // In a spawned process with an open-but-silent stdin, `isTTY` is undefined
+    // and `end` never fires (oclif's own parser defends against the same hang):
+    // give the FIRST byte a deadline, then let a slow-but-real pipe stream for
+    // as long as it needs.
+    const timeout = setTimeout(() => {
+      reject(
+        new Error(
+          `No data received on stdin after ${STDIN_FIRST_BYTE_TIMEOUT_MS / 1000}s. ` +
+            'Pass a JSON file argument, or pipe the patch via stdin.',
+        ),
+      );
+    }, STDIN_FIRST_BYTE_TIMEOUT_MS);
+
     process.stdin.setEncoding('utf8');
     process.stdin.on('data', chunk => {
+      clearTimeout(timeout);
       data += chunk;
     });
-    process.stdin.on('end', () => resolve(data));
-    process.stdin.on('error', reject);
+    process.stdin.on('end', () => {
+      clearTimeout(timeout);
+      resolve(data);
+    });
+    process.stdin.on('error', error => {
+      clearTimeout(timeout);
+      reject(error);
+    });
   });
 }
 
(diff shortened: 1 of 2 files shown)


Gawtier:
Re-verified this PR with @PMerlet's two commits. **CI is green** (build, lint, Macroscope, tests on Node 18/20/22), and I ran an **end-to-end test against a real throwaway demo project** (Development / Operations env). All of the added behaviour works:

- `--force` + confirmation prompt — aborts on `n`, sends on `y`
- stdin non-interactive guard — lists exactly the missing flags
- the "secret ≠ env" fix — with `FOREST_ENV_SECRET` set it still requires `-e/--env` (confirmed)
- real send via stdin + re-pull confirms the change landed; layout restored afterwards

**One bug found, pre-existing (from the original PR, not these commits):** piping without an explicit `-` — e.g. `echo '{...}' | forest layout patch --force -e … -t …`, which is exactly the form shown in the command's help/error examples — fails with `ENOENT`. oclif's parser consumes stdin to fill the optional positional `file` arg ([`parser/parse.js`](https://github.com/oclif/core/blob/main/src/parser/parse.ts) reads stdin when a positional is missing and `ignoreStdin` isn't set), so `args.file` becomes the piped JSON, `fromStdin` evaluates to `false`, we take the file branch and try to `readFile()` a filename made of the JSON. Side effect: the new `assertNonInteractiveFlags` guard never fires in that case. It fails loudly (no risk of a wrong patch being sent), but the documented invocation is broken and the error is confusing.

**Trivial fix:** add `ignoreStdin: true` to the `file` arg so oclif leaves stdin to our own reader (which also handles multi-line JSON correctly). Happy to push it to the branch — let me know.

PMerlet:
Thanks for the thorough e2e pass! Your diagnosis was spot on — verified against the installed `@oclif/core` parser: `parser/parse.js` fills any missing positional from stdin unless the arg sets `ignoreStdin`, so the piped JSON became `args.file`, `fromStdin` evaluated to `false`, and both the non-interactive guard and our own stdin reader were bypassed before the `ENOENT`.

Fixed in b932220 exactly as you suggested: `ignoreStdin: true` on the `file` arg (with a comment explaining why oclif must leave stdin alone). All three documented invocations now behave: pipe without `-`, pipe with `-`, and an explicit file path. Lint/tsc/unit tests green locally; CI is running on the commit.

christophebrun-forest (approved):
I've fixed the last macroscope comment.
LGTM

PR #801, opened 2026-07-05
feat(layout): round-trip workflow BPMN via `--with-workflows`

Gawtier:
## What

Opt-in `--with-workflows` on both sides of the layout-as-code loop:

- `layout pull --with-workflows` → downloads each workflow's stored BPMN into sidecar files `workflows/<id>.bpmn` next to `forest-layout.json`.
- `layout apply --with-workflows` → re-uploads those sidecars to the **target** env and relinks `bpmnAwsS3Identifier`.

This is the "pull + apply back" round-trip: **back up** a project's workflows, or **promote** them dev → prod, with faithful BPMN transport.

## Behaviour

- **Precedence**: a workflow authored with `steps` in the file is compiled and wins; sidecars only cover file workflows *without* `steps` (including UI-authored ones — the CLI has no `steps` for those, so verbatim transport is the only faithful option).
- **Dead refs are skipped, not fatal**: a `bpmnAwsS3Identifier` that no longer resolves to an S3 version (e.g. baked by a demo seed but never uploaded to this bucket → `NoSuchVersion`) is skipped with a warning instead of aborting the whole pull.

## Tested

Live on a demo project: `pull → apply back → pull` yields fresh S3 version ids **and** byte-identical BPMN (faithful round-trip). Dead refs from the bundled demo were skipped cleanly (4 live workflows transported, 9 dead refs warned).

> [!NOTE]
> ### Add `--with-workflows` flag to round-trip workflow BPMN sidecars in `layout pull` and `layout apply`
> - `layout pull --with-workflows` downloads each workflow's BPMN from S3 into `workflows/<id>.bpmn` sidecar files alongside the layout file, pruning stale sidecars for workflows no longer in the environment.
> - `layout apply --with-workflows` reads those sidecar files, uploads BPMN to the target environment, and strips non-portable `bpmnAwsS3Identifier` JSON patch ops so S3 pointers from the source env are never written directly.
> - Without `--with-workflows`, `layout apply` warns when it detects that applying would plant cross-environment S3 references.
> - New utilities in [`workflow-sidecar.ts`](https://github.com/ForestAdmin/toolbelt/pull/801/files#diff-cb2e3ee3186a5c71888642a9b83cd5e7332e80b5c6394436edf1a7430e9eebdd) handle id safety (path traversal guards), upload planning (skip identical bytes, resolve by id then name for promote scenarios), and stale sidecar selection.
> - Risk: sidecar uploads for workflows added in the same apply run require a re-fetch to resolve new workflow ids; if resolution fails the upload is skipped with a warning.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized be38184.</sup>
> 
>

Part of [PRD-715 — Workflows in CLI](https://linear.app/forestadmin/issue/PRD-715/workflows-in-cli).

### DIFF
--- a/src/commands/layout/apply.ts
+++ b/src/commands/layout/apply.ts
@@ -2,7 +2,7 @@ import type { LayoutScope } from '../../services/layout/types';
 import type { Config } from '@oclif/core';
 
 import { Args, Flags } from '@oclif/core';
-import { readFileSync } from 'fs';
+import { existsSync, readFileSync } from 'fs';
 import path from 'path';
 
 import AbstractAuthenticatedCommand from '../../abstract-authenticated-command';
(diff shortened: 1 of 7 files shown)


Gawtier:
Addressed all three Macroscope findings in `5bb75e2`: (1) **path traversal** — new `sidecarPath` rejects any traversal-unsafe workflow id (both pull-write and apply-read go through it); (2) **masked errors** — the pull catch now only skips a 404 (gone S3 version), so auth/network/disk errors propagate; (3) **stale sidecars** — pull now prunes `<id>.bpmn` files it did not re-write, so apply can't re-upload outdated BPMN.

Gawtier:
Addressed the two remaining Macroscope Mediums in `eff10e4`:
- **Idempotency** — `planSidecarUploads` now byte-compares each sidecar against the stored BPMN (like `planWorkflowBpmn`) and uploads only changed ones. Re-applying an unchanged file is a genuine no-op (verified: `pull → apply back` → *Nothing to apply*).
- **Non-portable ref** — `stripWorkflowBpmnOps` drops the file's `bpmnAwsS3Identifier` under `--with-workflows` (the upload owns that pointer), so a workflow without a checked-in sidecar can no longer be patched to the source env's nonexistent S3 version. Such workflows are now reported and left on the target's own BPMN.

PMerlet (approved):

PR #806, opened 2026-07-24
feat(projects): add `forest projects:create:in-app`

Gawtier:
## What

New command `forest projects:create:in-app` — register a Forest project for an app the user **hosts themselves** (Rails, Express, NestJS, …). Unlike `create:sql`/`demo`/`nosql`, it does **not** scaffold an agent and does **not** introspect a database: the project is created with `architecture: 'in-app'` and no agent (exactly like the web UI's in-app onboarding), and the command prints the env/auth secrets + how to mount Forest in the existing app. The user's own running agent pushes its schema on first boot.

- Makes `architecture` overridable on `AbstractProjectCreateCommand` (default stays `'microservice'`).
- In-app specifics: `requiresDatabase = false`, `agent = null` (so the server keeps `architecture: 'in-app'`), `dump` is a no-op.
- **Rails guidance lists the 5 required gems.** `gem 'forest_admin_rails'` alone installs but fails to boot (`cannot load such file -- forest_admin_agent`) because 1.36.0 doesn't declare its 4 companion gems as runtime deps — so the single-gem instruction would be wrong. (Worth a separate report to the Ruby SDK team.)

## Tested

- **Unit**: `in-app.test.js` (mirrors `demo.test.js`) — asserts the in-app create posts `architecture: 'in-app'` with no agent, skips all DB steps, scaffolds nothing, and prints the secrets + the 5-gem Rails guidance. `createProject` fixture gained an `architecture` param (defaults to `'microservice'`, backward-compatible).
  - _Updated in `0a8d5cd`_: 6 cases in `in-app.test.js` now, covering the `--format json` contract (no prompting, pure-JSON stdout, refusal when logged out, error visibility, no partial document), plus new unit tests for `Logger.reserveStdout` and `option-parser.getDefaultOptions`. See [the comment below](https://github.com/ForestAdmin/toolbelt/pull/806#issuecomment-5664570822) for what that commit changes and why.
- **Live end-to-end on BOTH stacks**, against real Forest projects created by this command, verified **server-side** (not just agent logs):
  - **Node** — Express + `@forestadmin/agent@1.90.3` mounted with the CLI-issued secret → dev env `is_active` false→true, **collections 0 → 8**.
  - **Ruby** — `rails new` (Rails 8.0.5 / Ruby 3.3.9) + the 5 forest gems + `rails g forest_admin_rails:install <secret>` → `is_active` false→true, **collections 0 → 2** (`Company`, `Customer`).
  - Test projects were torn down afterwards.

> Note: the `sql`/`nosql`/`create` suites require a live Postgres test DB (they fail locally with `SequelizeConnectionRefusedError`, same as on `main`); they pass on CI.

> [!NOTE]
> ### Add `projects:create:in-app` command with JSON mode and non-interactive defaults
> - Adds the `projects:create:in-app` command to register user-hosted applications with `in-app` architecture, skipping database and agent scaffolding.
> - Extends `AbstractProjectCreateCommand` with overridable `interactive` and `architecture` fields, retaining existing `microservice` and interactive defaults for other commands.
> - Updates `getCommandLineOptions` to accept an optional `interactive` parameter, resolving missing values from declared defaults without prompting when disabled.
> - Adds `reserveStdout` to `Logger` which routes normal log lines to stderr, keeping stdout free for JSON payloads in the new command.
> - Risk: `getCommandLineOptions` in [option-parser.ts](https://github.com/ForestAdmin/toolbelt/pull/806/files#diff-7f0c805f54f4277e7afc886ccd7eb30b23f9c09f792039769934a993ceca4849) changes signature to add `interactive`; `Logger` in [logger.js](https://github.com/ForestAdmin/toolbelt/pull/806/files#diff-6521ff6e75ade6e0e07c1614774bae0eb15e2fd8420eb0625fca8d3d32d067de) adds `reserveStdout` initialization. Existing callers retain default behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a8d5cd.</sup>
> 
>

### DIFF
--- a/src/abstract-project-create-command.ts
+++ b/src/abstract-project-create-command.ts
@@ -38,6 +38,11 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
   // test, introspection) are skipped.
   protected readonly requiresDatabase: boolean = true;
 
+  // Hosting architecture sent to the server. 'microservice' = a dedicated agent
+  // we scaffold (the default for every create:* command); 'in-app' = the user
+  // hosts the agent inside their own app (no scaffold).
+  protected readonly architecture: string = 'microservice';
+
   static override args = {
     applicationName: Args.string({
       name: 'applicationName',
@@ -190,7 +195,7 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
       agent:
         this.agent || (dbConfig.dbDialect === 'mongodb' ? 'express-mongoose' : 'express-sequelize'),
       dbDialect: dbConfig.dbDialect,
-      architecture: 'microservice',
+      architecture: this.architecture,
       isLocal: ['localhost', '127.0.0.1', '::1'].some(keyword =>
         dbConfig.dbHostname
           ? dbConfig.dbHostname.includes(keyword)
--- a/src/commands/projects/create/in-app.ts
+++ b/src/commands/projects/create/in-app.ts
@@ -1,0 +1,102 @@
+import type { Config } from '../../../interfaces/project-create-interface';
+import type { CommandOptions } from '../../../utils/option-parser';
+import type { Config as OclifConfig } from '@oclif/core';
+
+import AbstractProjectCreateCommand from '../../../abstract-project-create-command';
+import * as projectCreateOptions from '../../../services/projects/create/options';
+import { optionsToFlags } from '../../../utils/option-parser';
+
+/**
+ * Register a Forest Admin project for an app the user hosts themselves (in-app:
+ * Rails, Express, NestJS, …). Unlike create:sql/demo/nosql, this does NOT
+ * scaffold an agent and does NOT introspect a database — the schema is pushed by
+ * the user's own running agent on first boot. It just creates the project (with
+ * `architecture: 'in-app'`), and prints the environment secret to plug into the
+ * existing app.
+ */
+export default class InAppCommand extends AbstractProjectCreateCommand {
+  protected static options: CommandOptions = {
+    // The URL/port where the user's app runs — used for the dev environment endpoint.
+    applicationHost: projectCreateOptions.applicationHost,
+    applicationPort: projectCreateOptions.applicationPort,
+  };
+
+  /** @see https://oclif.io/docs/args */
+  static override readonly args = AbstractProjectCreateCommand.args;
+
+  /** @see https://oclif.io/docs/flags */
+  static override readonly flags = optionsToFlags(this.options);
+
+  static override readonly description =
+    'Register a project for an app you host yourself (in-app) and print its environment secret — no scaffold, no database.';
+
+  // In-app users bring their own agent (Rails gem, @forestadmin/agent, …), so no
+  // agent is sent (see getConfig): the server forces architecture=microservice
+  // whenever an agent is present, exactly as the web UI avoids by omitting it.
+  protected readonly agent = null;
+
+  // In-app: no DB introspection (the running agent pushes the schema itself)…
+  protected override readonly requiresDatabase = false;
+
+  // …and the project is hosted inside the user's app, not a scaffolded microservice.
+  protected override readonly architecture = 'in-app';
+
+  private forestEnvSecret?: string;
+
+  private forestAuthSecret?: string;
+
+  // Create the project WITHOUT an agent (like the web UI) so the server keeps
+  // architecture='in-app'. The abstract otherwise falls back agent → express-sequelize.
+  protected override async getConfig() {
+    const config = await super.getConfig();
+    config.meta.agent = null as unknown as string;
+
+    return config;
+  }
+
+  // Required by the abstract command, but in-app scaffolds nothing.
+  // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-empty-function -- intentional no-op
+  protected override async dump(): Promise<void> {}
+
+  // Skip file generation entirely; keep the secrets to surface in logNextSteps().
+  protected override async generateProject(config: Config): Promise<void> {
+    this.forestEnvSecret = config.forestEnvSecret;
+    this.forestAuthSecret = config.forestAuthSecret;
+  }
+
+  protected override logNextSteps(): void {
+    this.logger.info('In-app project created — no code was scaffolded.');
+    // Printed (not TTY-gated) on purpose: `npx forest-start` reads these back from stdout to wire
+    // the app, so gating on isTTY would break it. FOREST_AUTH_SECRET is a value you own (any random
+    // string works); FOREST_ENV_SECRET is the sensitive one.
+    this.logger.info('Set these on your app, then mount the Forest agent in your server:');
+    this.logger.info(`  FOREST_ENV_SECRET=${this.forestEnvSecret}`);
+    this.logger.info(
+      `  FOREST_AUTH_SECRET=${this.forestAuthSecret}   (you own this one — keep it or set your own)`,
+    );
+    this.logger.info(
+      this.chalk.yellow(
+        '  ⚠ These are secrets — keep them out of version control and shared CI logs.',
+      ),
+    );
+    // The 5 gems are all required: forest_admin_rails does not declare its companions as
+    // runtime deps, so `gem 'forest_admin_rails'` alone installs but fails to boot.
+    this.logger.info(
+      `Rails:   add the 5 Forest gems to your Gemfile — ${this.chalk.bold(
+        'forest_admin_rails, forest_admin_agent, forest_admin_datasource_toolkit, ' +
+          'forest_admin_datasource_customizer, forest_admin_datasource_active_record',
+      )},`,
+    );
+    this.logger.info(
+      `         then ${this.chalk.bold('bundle install')} and ${this.chalk.bold(
+        'rails g forest_admin_rails:install $FOREST_ENV_SECRET',
+      )}`,
+    );
+    this.logger.info(
+      `Node.js: ${this.chalk.bold(
+        'npm install @forestadmin/agent',
+      )} + a datasource package, then mount it on your server (createAgent(...).mountOnExpress(app).start()).`,
+    );
+    this.logger.info('On first boot your agent pushes its schema and your admin panel goes live.');
+  }
+}
--- a/src/commands/projects/create/in-app.ts
+++ b/src/commands/projects/create/in-app.ts
@@ -1,6 +1,7 @@
 import type { Config } from '../../../interfaces/project-create-interface';
 import type { CommandOptions } from '../../../utils/option-parser';
-import type { Config as OclifConfig } from '@oclif/core';
+
+import { Flags } from '@oclif/core';
 
 import AbstractProjectCreateCommand from '../../../abstract-project-create-command';
 import * as projectCreateOptions from '../../../services/projects/create/options';
@@ -18,14 +19,30 @@ export default class InAppCommand extends AbstractProjectCreateCommand {
   protected static options: CommandOptions = {
     // The URL/port where the user's app runs — used for the dev environment endpoint.
     applicationHost: projectCreateOptions.applicationHost,
-    applicationPort: projectCreateOptions.applicationPort,
+    applicationPort: {
+      ...projectCreateOptions.applicationPort,
+      // 3310 is where scaffolded microservice agents run; here the agent lives in
+      // the user's own app (Rails/Express/NestJS, …), which typically runs on 3000.
+      default: '3000',
+      // …and that app already exists, hence "is running" (not "will be").
+      prompter: { question: "What's the port on which your application is running?" },
+    },
   };
 
   /** @see https://oclif.io/docs/args */
   static override readonly args = AbstractProjectCreateCommand.args;
 
   /** @see https://oclif.io/docs/flags */
-  static override readonly flags = optionsToFlags(this.options);
+  static override readonly flags = {
+    ...optionsToFlags(this.options),
+    format: Flags.string({
+      description:
+        'Output format. With `json`, stdout carries only a machine-readable ' +
+        '`{"projectId", "envSecret", "authSecret"}` document (e.g. for `npx forest-start`).',
+      options: ['text', 'json'],
+      default: 'text',
+    }),
+  };
 
   static override readonly description =
     'Register a project for an app you host yourself (in-app) and print its environment secret — no scaffold, no database.';
@@ -45,6 +62,8 @@ export default class InAppCommand extends AbstractProjectCreateCommand {
 
   private forestAuthSecret?: string;
 
+  private jsonOutput = false;
+
   // Create the project WITHOUT an agent (like the web UI) so the server keeps
   // architecture='in-app'. The abstract otherwise falls back agent → express-sequelize.
   protected override async getConfig() {
@@ -52,6 +71,20 @@ export default class InAppCommand extends AbstractProjectCreateCommand {
     config.meta.agent = null as unknown as string;
 
     return config;
+  }
+
+  protected override async getCommandOptions(): Promise<projectCreateOptions.ProjectCreateOptions> {
+    const options = await super.getCommandOptions();
+
+    this.jsonOutput =
+      (options as projectCreateOptions.ProjectCreateOptions & { format?: string }).format ===
+      'json';
+    // In JSON mode stdout must carry ONLY the JSON document printed by
+    // logNextSteps(): mute the logger's stdout lines (spinners already write to
+    // stderr, and errors keep going to stderr).
+    if (this.jsonOutput) this.logger.silentStdout = true;
+
+    return options;
   }
 
   // Required by the abstract command, but in-app scaffolds nothing.
@@ -65,10 +98,27 @@ export default class InAppCommand extends AbstractProjectCreateCommand {
   }
 
   protected override logNextSteps(): void {
+    if (this.jsonOutput) {
+      // Machine-readable contract (what `npx forest-start` parses): stdout carries
+      // ONLY this JSON document — the logger's stdout is muted in this mode and
+      // progress spinners write to stderr — so callers can JSON.parse the whole output.
+      this.context.stdout.write(
+        `${JSON.stringify({
+          // The abstract command stores the created project id on the (singleton)
+          // eventSender meta; read it back from there rather than refactoring it.
+          projectId: this.context.eventSender.meta.projectId,
+          envSecret: this.forestEnvSecret,
+          authSecret: this.forestAuthSecret,
+        })}\n`,
+      );
+
+      return;
+    }
+
     this.logger.info('In-app project created — no code was scaffolded.');
-    // Printed (not TTY-gated) on purpose: `npx forest-start` reads these back from stdout to wire
-    // the app, so gating on isTTY would break it. FOREST_AUTH_SECRET is a value you own (any random
-    // string works); FOREST_ENV_SECRET is the sensitive one.
+    // Human-oriented output: scripts should use `--format json` instead (stable,
+    // parsable). FOREST_AUTH_SECRET is a value you own (any random string works);
+    // FOREST_ENV_SECRET is the sensitive one.
     this.logger.info('Set these on your app, then mount the Forest agent in your server:');
     this.logger.info(`  FOREST_ENV_SECRET=${this.forestEnvSecret}`);
     this.logger.info(
--- a/src/services/logger.js
+++ b/src/services/logger.js
@@ -15,6 +15,10 @@ class Logger {
 
     // FIXME: Silent was not used before as no "silent" value was in context.
     this.silent = !!this.env.SILENT && this.env.SILENT !== '0';
+
+    // When true, mutes stdout lines only (stderr still goes through). Used by
+    // commands whose stdout must stay machine-readable (e.g. `--format json`).
+    this.silentStdout = false;
   }
 
   _logLine(message, options) {
@@ -24,6 +28,8 @@ class Logger {
       ...DEFAULT_OPTION_VALUES,
       ...options,
     };
+
+    if (this.silentStdout && options.std !== 'err') return;
 
     let actualPrefix = '';
     if ([undefined, null, ''].indexOf(options.prefix) === -1) actualPrefix = `${options.prefix} `;
--- a/test/commands/projects/create/in-app.test.js
+++ b/test/commands/projects/create/in-app.test.js
@@ -1,0 +1,56 @@
+const testCli = require('../../test-cli-helper/test-cli');
+const InAppCommand = require('../../../../src/commands/projects/create/in-app').default;
+const { testEnvWithSecret } = require('../../../fixtures/env');
+const { createProject, updateNewEnvironmentEndpoint } = require('../../../fixtures/api');
+
+describe('projects:create:in-app', () => {
+  describe('execution', () => {
+    it('should register an in-app project (no agent, no DB, no scaffold) and print the secrets + guidance', () =>
+      testCli({
+        commandClass: InAppCommand,
+        commandArgs: ['name'],
+        env: testEnvWithSecret,
+        token: 'any',
+        // Only host + port are prompted (no `language`: nothing is scaffolded).
+        prompts: [
+          {
+            in: [
+              {
+                name: 'applicationHost',
+                message: "What's the IP/hostname on which your application will be running?",
+                type: 'input',
+                default: 'http://localhost',
+                validate: expect.any(Function),
+              },
+              {
+                name: 'applicationPort',
+                message: "What's the port on which your application will be running?",
+                type: 'input',
+                default: '3310',
+                validate: expect.any(Function),
+              },
+            ],
+            out: { applicationHost: 'http://localhost', applicationPort: '3310' },
+          },
+        ],
+        api: [
+          // The whole point: architecture 'in-app' + no agent (unlike every other create:*).
+          () => createProject({ databaseType: null, agent: null, architecture: 'in-app' }),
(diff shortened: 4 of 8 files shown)


PMerlet:
Pushed `0a8d5cd` on top of `3309e8b` (fast-forward, nothing of yours discarded). Context below, since it touches files beyond the new command.

### Why

`--format json` documents a contract — *"stdout carries only a machine-readable document"* — that it did not keep. Three ways, all reproduced on the built binary:

1. **No JSON at all in the normal scripted case.** `applicationHost`/`applicationPort` have prompters and their defaults are only applied *by* inquirer, so they were always prompted. `forest projects:create:in-app x --format json < /dev/null` wrote an inquirer question plus ANSI escapes to stdout and **exited 0**. The tests missed it because inquirer is mocked.
2. **Login leaked to stdout.** `checkAuthentication()` runs before `getCommandOptions()`, so the mute was installed too late: an unauthenticated run wrote `Login required.` and a password prompt to stdout.
3. **Errors silently swallowed.** The real cause is logged with `logger.log()`, which the mute dropped. Text mode printed `Error: connect ECONNREFUSED …`; JSON mode printed nothing about the cause.

### What changed

- `Logger.silentStdout` → `Logger.reserveStdout`, which **diverts** stdout lines to stderr instead of dropping them. Nothing is lost — diagnostics stay visible and stop polluting the parsable stream. This is what fixes (3).
- `option-parser` gains `getDefaultOptions()`, an unattended counterpart to `getInteractiveOptions()` using each option's declared default, selected via `getCommandLineOptions(instance, { interactive })`.
- `AbstractProjectCreateCommand` exposes an `interactive` getter as the extension point; `in-app` returns `false` in JSON mode.
- `in-app` parses `--format` in `run()`, before the login check, and refuses with exit 10 rather than hanging on a password prompt.

### Two things worth your attention

**Shared files.** `Logger`, `option-parser`, `AbstractProjectCreateCommand` and the test harness are touched. The changes are additive and default to current behaviour (`interactive` defaults to `true`, `reserveStdout` to `false`), and `getCommandLineOptions` has no other caller — but it goes beyond `in-app`, so it deserves a look.

**One out-of-scope addition.** While exercising the fix I found that a missing env secret surfaced as *success*: `JSON.stringify` drops undefined keys, so JSON mode emitted `{"projectId":…}` with exit 0, and text mode printed `FOREST_ENV_SECRET=undefined`. `generateProject()` now refuses both. Isolated and tested — easy to drop if you want the PR minimal. Note the base class still prefixes `Cannot generate your project.` when that guard fires even though the project *was* created; the next line corrects it. Fixing that message would touch every `create:*`, so I left it alone.

### Verified

CI 6/6. Full suite locally: **114 failures / 1131** against **114 / 1119** on `3309e8b` — same pre-existing Postgres failures, +12 tests. Each fix was mutation-tested (reverted individually) to confirm the new tests actually fail without it.

On the built binary, against a local stub API:

| scenario | before | after |
| --- | --- | --- |
| json, no flags, stdin closed | exit 0, prompt + ANSI on stdout | exit 0, pure JSON, `:3000` default applied |
| json, server unreachable | exit 0, cause hidden | exit 1, stdout empty, `ECONNREFUSED` on stderr |
| json, logged out | prompt on stdout | exit 10, stdout empty |
| json, update available | — | oclif warning goes to stderr, stdout stays parsable |

### Rails gems

Left exactly as you wrote them. I re-verified your finding and it still holds on **1.43.0** (published 4 days ago, 7 minors after the 1.36.0 you tested): `lib/forest_admin_rails.rb:6` requires `forest_admin_agent`, and the gemspec declares 11 runtime deps, none of them a `forest_*` gem. Filed as PRD-1223 so the CLI workaround can be removed once the gemspec is fixed.

PMerlet (approved):

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+  // In-app users bring their own agent (Rails gem, @forestadmin/agent, …), so no
+  // agent is sent (see getConfig): the server forces architecture=microservice
+  // whenever an agent is present, exactly as the web UI avoids by omitting it.
+  protected readonly agent = null;
+
+  // In-app: no DB introspection (the running agent pushes the schema itself)…
+  protected override readonly requiresDatabase = false;
+
+  // …and the project is hosted inside the user's app, not a scaffolded microservice.
+  protected override readonly architecture = 'in-app';
```
PMerlet, line 59: **Applies to:** `src/commands/init.js:87`, not in this diff
Claude Opus 5 (claude-opus-5): **Should fix**: A user who runs `forest init` on a project this command created is asked to configure a database.

`init.js:87` branches on `origin` equal to `In-app`, which `project-creator.js:49` never sends for this new architecture.

A way out: read `origin` off a project your live run created, and send it from `project-creator.js` if it is `Lumber`.
</details>

1. Run `forest projects:create:in-app myapp`.
2. Run `forest init` and select that project.
3. It asks you to configure a database the in-app agent already owns.
</details>

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+      // eventSender meta; read it back from there rather than refactoring it.
+      const { projectId } = this.context.eventSender.meta;
+
+      // Machine-readable contract (what `npx forest-start` parses): stdout carries
+      // ONLY this JSON document — logger lines are diverted to stderr, spinners
+      // already write there, and nothing prompts — so callers can JSON.parse the
+      // whole of stdout.
+      this.context.stdout.write(
+        `${JSON.stringify({
+          projectId,
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: A caller parsing `--format json` gets `projectId` as a string, while the contract test locks a number.

This writes the deserialized `id` uncoerced, and `jsonapi-serializer` passes a compliant string `id` straight through.

A way out: coerce the id to a string here, or fix `test/fixtures/api.js:1537` so the test locks the real shape.
</details>

1. Point the CLI at a JSON:API compliant Forest server.
2. Run `forest projects:create:in-app myapp --format json`.
3. stdout carries `"projectId":"4242"`, not the `4242` the new test asserts.
</details>

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+
+      return;
+    }
+
+    this.logger.info('In-app project created — no code was scaffolded.');
+    // Human-oriented output: scripts should use `--format json` instead (stable,
+    // parsable). FOREST_AUTH_SECRET is a value you own (any random string works);
+    // FOREST_ENV_SECRET is the sensitive one.
+    this.logger.info('Set these on your app, then mount the Forest agent in your server:');
+    this.logger.info(`  FOREST_ENV_SECRET=${this.forestEnvSecret}`);
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: With `SILENT` set, the command creates the project, exits 0, and never shows the environment secret.

`logger.js:27` drops every line when `SILENT` is set, and this is the only delivery of `FOREST_ENV_SECRET` in text mode.

A way out: write the secrets through `this.context.stdout` the way the JSON branch already does.
</details>

1. Export `SILENT=1`.
2. Run `forest projects:create:in-app myapp`.
3. The project exists in Forest and the terminal printed nothing at all.
</details>

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (this.jsonOutput) {
+      // The abstract command stores the created project id on the (singleton)
+      // eventSender meta; read it back from there rather than refactoring it.
+      const { projectId } = this.context.eventSender.meta;
+
+      // Machine-readable contract (what `npx forest-start` parses): stdout carries
+      // ONLY this JSON document — logger lines are diverted to stderr, spinners
+      // already write there, and nothing prompts — so callers can JSON.parse the
+      // whole of stdout.
+      this.context.stdout.write(
```
PMerlet, line 151: Claude Opus 5 (claude-opus-5): **Should fix**: The JSON document omits the endpoint that was registered, so a wrong default is undetectable by the caller.

`api.js:106` builds the dev environment endpoint from the silently defaulted host and port, which this write omits.

A way out: add the resolved endpoint to the document, or require `-H` and `-P` under `--format json`.
</details>

1. Run `forest projects:create:in-app myapp --format json` with no `-H` and no `-P`.
2. The document parses cleanly and the exit code is 0.
3. The dev environment points at `http://localhost:3000` and the admin panel stays dead.
</details>

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    // The secrets ARE the deliverable here — nothing was scaffolded to fall back on.
+    // Checked before the success path runs, so we never claim success and then print
+    // `FOREST_ENV_SECRET=undefined` (text) or drop the key altogether (json, where
+    // JSON.stringify omits undefined and the caller would see a valid-looking
+    // document with exit 0).
+    if (!this.forestEnvSecret || !this.forestAuthSecret) {
+      const { projectId } = this.context.eventSender.meta;
+
+      throw new Error(
```
PMerlet, line 133: **Applies to:** `src/abstract-project-create-command.ts:129`, not in this diff
Claude Opus 5 (claude-opus-5): **Should fix**: The failure path tells the operator the project could not be created, after it was created.

`abstract-project-create-command.ts:129` prints `Cannot generate your project.` in red above the truthful line this throw carries.

A way out: make that headline conditional on `eventSender.meta.projectId`, and spread its two strings instead of passing an array.
</details>

1. Create a project whose environment PUT returns no `secretKey`, as the new test does.
2. Read stderr, whose first line is the red contradiction rendered as a raw JSON array.
3. Re-run the command, which silently re-attaches to the project already created.
</details>

src/commands/projects/create/in-app.ts
```diff
@@ -0,0 +1,192 @@
...
+  /** @see https://oclif.io/docs/args */
+  static override readonly args = AbstractProjectCreateCommand.args;
+
+  /** @see https://oclif.io/docs/flags */
+  static override readonly flags = {
+    ...optionsToFlags(this.options),
+    format: Flags.string({
+      description:
+        'Output format. With `json`, stdout carries only a machine-readable ' +
+        '`{"projectId", "envSecret", "authSecret"}` document (e.g. for `npx forest-start`).',
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: `--help` points the user at `npx forest-start`, which is not published on npm.

This string sits in the oclif flag description that `--help` prints, and `npm view forest-start` returns 404.

A way out: drop `npx forest-start` from this description until the consumer ships.
</details>

1. Run `forest projects:create:in-app --help`.
2. Read the `--format` description.
3. Run `npx forest-start` and get a 404 from the registry.
</details>

test/services/logger.unit.test.js
```diff
@@ -51,6 +51,49 @@ describe('services > Logger', () => {
...
+        expect.assertions(2);
+        const context = makeContext();
+        const { stdout, stderr } = context;
+
+        const logger = new Logger(context);
+        logger.reserveStdout = true;
+        logger.log('an error', { std: 'err' });
+
+        expect(stdout.write).toHaveBeenCalledTimes(0);
+        expect(stderr.write).toHaveBeenCalledTimes(1);
```
PMerlet, line 79: Claude Opus 5 (claude-opus-5): **Violates conventions**: The stderr assertion stops at called-ness, so a mangled message still passes.

This asserts only `toHaveBeenCalledTimes`, which `testing.md#Assert behavior with args and outputs` bans for a writable argument.

A way out: assert the expected line with `toHaveBeenCalledWith`, as the sibling test at line 66 does.
</details>

PMerlet (commented):
**Spec:** no Linear ticket is linked from this PR, so the functional check was limited to the PR description.

**Applies to:** the PR as a whole
Claude Opus 5 (claude-opus-5): **Should fix**: The in-app architecture is expressed by omitting the agent, and no ADR records that choice.

`in-app.ts:107` nulls `agent` so the server infers `in-app`, and the ADR search found nothing covering that coupling.

A way out: run `/adr` to record why an absence encodes the architecture, and how it relates to `origin`.
</details>

1. Delete `config.meta.agent = null` from `getConfig`.
2. Run `forest projects:create:in-app myapp`.
3. The project is recorded as `microservice`, with no error from either side.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): Posted here in error by a misindexed loop, this text answers the `projectId` finding and has been moved to its own thread. The reply below is this thread's answer.

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): `projectId` is now pinned to a string in `in-app.ts`, and the fixture serves the compliant id a real server sends.

29c1fe2

Both halves, because either alone leaves a gap.

The coercion pins the contract whatever the server returns, and the fixture stops the test locking a shape no compliant server sends.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): The two secret lines now go straight to stdout in `logNextSteps`, so `SILENT` no longer drops them.

29c1fe2

Only the secrets moved, and the surrounding guidance stays on the logger, which is what `SILENT` is for.

A new test asserts both secrets survive `SILENT` while the guidance goes quiet.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): The json document now carries the endpoint the API registered, threaded back from `project-creator` rather than recomputed.

29c1fe2

Reporting what was registered rather than rebuilding it from host and port keeps one source of truth, since `api.js:106` owns that construction.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): The headline is now conditional on a created project, and its two strings are spread instead of passed as an array.

29c1fe2

Guarded with `meta?.projectId`, because a parse failure aborts before `meta` is assigned and an unguarded read changed the exit code.

The three tests that locked the raw array now assert the two lines.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): Dropped `npx forest-start` from the format flag description, which now names the document keys only.

29c1fe2

src/commands/projects/create/in-app.ts
PMerlet: Claude Opus 5 (claude-opus-5): The claim does not hold, the server stores origin In-app for an in-app project, so `forest init` already skips the database step.

Confirmed with the author against the live API: origin is derived from the architecture, not from the creating client, so `init.js:87` takes the In-app branch.

Nothing was available to fetch either way, since `serializers/project.js` carries no connection string and the URL never leaves the machine.
</details>

PMerlet:
Claude Opus 5 (claude-opus-5): No ADR was written, the constraint is recorded in `getConfig` where the null agent is set.

29c1fe2

An ADR records a decision with alternatives, and this is a server contract the command obeys.

A non-null agent forces architecture back to microservice, so both halves are required and no alternative exists.

The load-bearing null now says so at its site, which is where a reader meets it.
</details>

test/services/logger.unit.test.js
PMerlet: Claude Opus 5 (claude-opus-5): The stderr assertion now uses `toHaveBeenCalledWith` on the expected line, matching the sibling test above it.

29c1fe2

src/commands/projects/create/in-app.ts
PMerlet: Accepted, the counter-argument stands and the finding was wrong.
<details><summary>How it was checked</summary>

Predicate: a project created by this command returns an origin other than In-app, sending `init.js` down the database branch.

The premise is server side and unreadable here, so the fact was confirmed with the author in session.

Nothing pinned since no code changed, and the counter-argument is recorded in a comment on the pull request.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Closed, the emitted type no longer varies with the server, and the regression is unpinned.
<details><summary>How it was checked</summary>

Predicate: a defined project id reaches stdout as something other than a string.

Ran the real deserializer and the emitting expression over number and string ids, and String is total over any defined id.

Not pinned, because the compliant fixture alone makes the new assertion pass on the uncoerced code.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Closed, both secrets now survive SILENT, and a test pins it.
<details><summary>How it was checked</summary>

Predicate: a successful run with SILENT set emits neither secret on any stream.

The secrets bypass the logger entirely, so the silent guard in `_logLine` cannot reach them.

Pinned by a new test asserting both secrets under SILENT, which fails on the logger based code.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Closed, the document now carries the registered endpoint, and two tests pin it.
<details><summary>How it was checked</summary>

Predicate: a json run emits a document without the endpoint the API registered.

Swept the single production caller and the only reader, and `api.js` sets `apiEndpoint` unconditionally on the created and refetched paths.

Pinned by two json tests that strict compare the whole document, which lacked the key before.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Closed for its own predicate, and the shared guard cost the other commands a message.
<details><summary>How it was checked</summary>

Predicate: a failure after the project exists still prints the headline, or prints it as an array.

Neither reproduces, but the sweep of all seven subclasses found the guard also suppresses the community guidance after creation.

Not pinned, and that suppression is raised as a separate finding on its own thread.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Closed, no user facing string names the package, and the regression is unpinned.
<details><summary>How it was checked</summary>

Predicate: a string that reaches help output names `npx forest-start`.

Grep over the shipped source at this head finds it only in two internal comments, which are stale but never printed.

Not pinned, because asserting help text in a test is brittle and nothing here does it.
</details>

test/services/logger.unit.test.js
PMerlet: Closed, every written argument in that block is now asserted.
<details><summary>How it was checked</summary>

Held against the cited section, which bans stopping at called-ness and exempts asserting a spy was not called.

The two writing tests now carry `toHaveBeenCalledWith`, and the remaining bare counts are all zero call assertions.

No pin, since a convention has no regression to pin.
</details>

PMerlet:
Closed, the coupling is now recorded at the site that sets the null agent.
<details><summary>How it was checked</summary>

Predicate: the architecture depends on an absent agent and nothing in the repo records why.

`getConfig` now states that a non-null agent makes the server force architecture back to microservice.

Not pinned, and the case against an ADR stands because the constraint is the server's with no alternative.
</details>

PMerlet:
Claude Opus 5 (claude-opus-5): Recorded outside its thread, as `/verify-fixes` requires for an accepted counter-argument.

The finding that `forest init` would ask an in-app project for a database does not hold. The server stores `origin: 'In-app'` for a project created with `architecture: 'in-app'`, so `init.js:87` takes the In-app branch and skips both the database prompt and the `.env` prompt. Confirmed with the author against the live API, since the fact is server side and unreadable from this repo.

src/abstract-project-create-command.ts
```diff
@@ -117,7 +138,13 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
       }
       // Display customized error for non-authentication errors.
       else if (error.status !== 401 && error.status !== 403) {
-        this.logger.error(['Cannot generate your project.', `${this.messages.ERROR_UNEXPECTED}`]);
+        // Once the project exists this headline is false, and the error thrown from
+        // here on carries the truthful line instead. Passing an array would print it
+        // as a raw JSON array, since the logger stringifies a non-string message.
+        if (!this.eventSender.meta?.projectId) {
```
PMerlet, line 144: Claude Opus 5 (claude-opus-5): **Should fix**: A scaffolding command that fails after creating the project now prints no pointer to where help lives.

This guard suppresses `ERROR_UNEXPECTED` for all seven subclasses, though only in-app throws an error explicit enough to stand alone.

A way out: keep the guidance line unconditional, and gate only the `Cannot generate your project.` headline.
</details>

1. Run `forest projects:create:sql` with a database password that is wrong.
2. The project is created on Forest, then the connection test fails.
3. stderr carries the raw error alone, without the community link every other failure prints.
</details>

PMerlet: Claude Opus 5 (claude-opus-5): The guidance line is now unconditional in `abstract-project-create-command.ts`, and only the headline stays gated on a created project.

8fc8b24

Fixed at the class rather than the cited command, since the guard sits in the shared abstract and reached all seven subclasses.

Pinned at two callers, `nosql` for a scaffolding command and `in-app`, both asserting the guidance survives and the headline does not.

Confirmed by reverting the source hunk and watching both fail, so the pin holds rather than merely existing.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still accepted, and nothing in the new commit touches the premise.
<details><summary>How it was checked</summary>

Predicate: a project created by this command returns an origin other than In-app, sending `init.js` down the database branch.

The new commit touches only the shared catch block and two test files, so the server side premise is unmoved.

Nothing pinned since no code changed, and the counter-argument stays recorded on the pull request.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still closed, the emitted type is unchanged and the regression is still unpinned.
<details><summary>How it was checked</summary>

Predicate: a defined project id reaches stdout as something other than a string.

The new commit leaves `in-app.ts` and the fixture untouched, so the executed result from the previous head still holds.

Not pinned, because the compliant fixture alone makes the assertion pass on the uncoerced code.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still closed, the SILENT test is untouched and still passes at this head.
<details><summary>How it was checked</summary>

Predicate: a successful run with SILENT set emits neither secret on any stream.

The new commit adds assertions to a different test, leaving this one and the stdout writes unchanged.

Pinned by the SILENT test, which passed in the full run at this head.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still closed, the destructure and the endpoint field survive the new commit intact.
<details><summary>How it was checked</summary>

Predicate: a json run emits a document without the endpoint the API registered.

Re-read the catch block's neighbourhood at this head, where the endpoint destructure and the assignment are unchanged.

Pinned by two json tests that strict compare the whole document, both passing at this head.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still closed, and now pinned, since the new commit added the assertion this verdict said was missing.
<details><summary>How it was checked</summary>

Predicate: a failure after the project exists still prints the headline, or prints it as an array.

Re-derived because the new commit moved this premise, and the headline is still gated and still a single string.

Pinned now by the in-app test asserting the headline is absent once the project exists.
</details>

src/commands/projects/create/in-app.ts
PMerlet: Still closed, no user facing string names the package at this head.
<details><summary>How it was checked</summary>

Predicate: a string that reaches help output names `npx forest-start`.

The new commit does not touch `in-app.ts`, so the grep result from the previous head still stands.

Not pinned, because asserting help text in a test is brittle and nothing here does it.
</details>

test/services/logger.unit.test.js
PMerlet: Still closed, the logger test file is untouched by the new commit.
<details><summary>How it was checked</summary>

Predicate: a spy assertion on a written argument stops at bare called-ness.

Held against the cited section again, and the new commit changes no file in that suite.

No pin, since a convention has no regression to pin.
</details>

src/abstract-project-create-command.ts
PMerlet: Closed, the guidance line is now outside every guard in that branch.
<details><summary>How it was checked</summary>

Predicate: a failure reaching this catch prints no pointer to where help lives.

Swept the five subclasses that reach it, since the two bundle commands override runAuthenticated and create no project.

Pinned at nosql and in-app, both asserting the guidance after creation, and both passing here.
</details>

PMerlet:
Still closed, the comment in `getConfig` is untouched by the new commit.
<details><summary>How it was checked</summary>

Predicate: the architecture depends on an absent agent and nothing in the repo records why.

The new commit touches only the shared catch block and two test files, leaving `getConfig` unchanged.

Not pinned, because a comment carries no behaviour a test could hold.
</details>

PMerlet (approved):
Approved after /verify-fixes run 806-20260922145351-PMerlet-verify: 9 findings of 806-20260921100148-PMerlet closed or accepted, none reopened.

PR #807, opened 2026-07-26
feat(projects): offer a connection URL in create:sql / create:nosql prompts

Gawtier:
## What

Interactive `projects:create:sql` / `projects:create:nosql` now open with an optional **connection URL prompt** — `Database connection URL (leave blank to enter the details manually):` (MongoDB wording on nosql). Paste a URL and the field prompts (dialect/name/host/port/user/password, Mongo SRV) are skipped; leave it blank and today's field-by-field flow runs unchanged.

This brings the CLI to parity with the web onboarding, which offers the same Form/URL choice. Until now the connection URL was reachable only via the `-c` flag (`prompter: null`), so a human running the interactive create could never paste a connection string.

## How it works

Scoped to the two commands — the shared options are untouched, so `init` / `projects:create` / `demo` prompts are unchanged:

- Each command overrides `databaseConnectionURL` with a leading `prompter` and a per-engine validator (`validateSqlConnectionUrl` / `validateMongoConnectionUrl` in `src/services/projects/create/options.ts`). No new flag, no extra "choose a method" question.
- The field prompts get `when: skipWhenConnectionUrl`, so they are skipped once a URL is answered.
- The URL option carries an `exclusive` list mirroring the field options': passing any DB field flag (`-d`/`-n`/`-h`/`-p`/`-u`/`--databasePassword`, `--mongoDBSRV`) drops the URL question, so scripted/non-TTY usage never faces a new question to hang on.
- The prompt is **masked** (`prompter.secret`, new): the URL embeds the database password, and `optionToInquirer` only masks options whose *name* matches `/(password|secret)/`. Without it the credentials were echoed on screen and left in the terminal scrollback — a regression against the field-by-field flow, whose `databasePassword` prompt is masked. A `*` mask keeps a failed paste distinguishable from the blank answer.
- The answer is **trimmed** at the source (`filter`, new — inquirer runs it before `when`/`validate`). A pasted URL routinely drags whitespace along, and the untrimmed value reaches `getDialect()` *during* prompting, so a leading space used to make the database schema question disappear. `getCommandOptions` normalizes the `-c` flag path the same way, and a blank answer becomes `undefined` so the DB config falls back to the field values.

## Behaviour change for scripted usage (`-c`)

`getCommandLineOptions` runs an option's `validate` on values coming from the command line too, so **the new validators also gate the `-c` flag**. `projects:create:sql -c mariadb://…` now fails at parse time where it used to be accepted.

This is deliberate. `runAuthenticated` creates the project on Forest *before* testing the database connection, so a URL the toolbelt cannot interpret either leaves an orphan project behind, or — worse — succeeds all the way and hands the user a generated project whose `package.json` ships no driver, broken at the first `npm start`. Failing in the parser, with a message that says what to do, is the point.

What is rejected, and why:

- `mariadb://` — `getDialect()` maps it to the `mysql` dialect, so the generated project ships `mysql2`, while `@forestadmin/datasource-sql` derives `mariadb` from the URL scheme and needs the `mariadb` driver. Use `mysql://`. (The dumper does have mariadb support, but `getDialect()` never returns that dialect; making it reachable is a separate change — `dbDialect` flows into the project meta sent to the API, whose v1 dialect choices only know `mysql`.)
- Anything outside `postgres://`, `postgresql://`, `mysql://`, `mssql://` (SQL) and `mongodb://`, `mongodb+srv://` (Mongo) — `pg://`, `mysql2://`, `sqlite://` and friends all end up with no dialect, hence no driver in the generated project.

What is **not** rejected: uppercase schemes. `getDialect()` was matching them case-sensitively and returning `null`, which is what actually broke `Postgres://…`. Both Sequelize and `@forestadmin/datasource-sql` parse the URL with a URL parser that normalizes the scheme, so `Postgres://user:MyPass@host/db` connects and generates a working project — `getDialect()` is now case-insensitive (on a lowercased copy: the credentials the URL carries are case-sensitive), in `src/services/projects/create/options.ts` and in `src/services/schema/update/database.js`, which had the same bug on the `schema:update` path. The mongo validator is the exception and still requires lowercase, because the mongodb driver itself throws `Invalid scheme, expected connection string to start with "mongodb://"`.

`projects:create` (v1, forest-express) is deliberately left untouched: adding a new hard failure to a legacy path costs more than the consistency is worth.

## Error messages

Every rejection used to read *"the scheme must be lowercase"*, including `mariadb://`, `pg://`, `sqlite://` and a URL with a leading space. Each case now says what is actually wrong: unsupported scheme (listing the supported ones), URL truncated after `://`, not a URL at all, and `mariadb://` pointing at `mysql://`.

## Tested

- `options.test.ts` — the helpers: skip predicate (incl. blank), scheme validation per engine, each exact message, uppercase accepted on SQL and refused on Mongo, `getDialect()` case-insensitivity leaving the credentials alone, and a guard against a scheme resolving on `Object.prototype`.
- `connection-url-prompt.test.ts` (new) — the command test helper stubs inquirer and returns canned answers, so nothing proved the field prompts were actually skipped. This one drives the **real** inquirer against the real command options and asserts which questions the user is asked (URL → fields skipped, blank → fields asked, nosql likewise), that the pasted credentials never reach the screen, and that the URL is stored trimmed.
- `database.unit.test.js` — `getDialect()` on the `schema:update` path, case-insensitive and non-destructive.
- `abstractProjectCreateCommand.unit.test.ts` — a blank URL normalizes to `undefined` and the connection uses the field values; a URL pasted with whitespace is trimmed before the dialect is derived.
- `sql.test.js` / `nosql.test.js` — interactive URL path (project created, dialect derived from the URL), field-flags path (no URL question in the prompt list), and the blank-URL round-trip against the test database.
- `tsc` + lint clean.

> [!NOTE]
> ### Add connection URL prompt to `create:sql` and `create:nosql` project commands
> - Adds a masked, validated connection URL prompt as the first option in both `create:sql` and `create:nosql`; when a URL is supplied, the individual database field prompts (host, port, user, password, name, dialect) are skipped
> - Adds shared validators ([options.ts](https://github.com/ForestAdmin/toolbelt/pull/807/files#diff-e48af1ebad8eb6d09ed721002096b6f9656989becf4358c0e3bab990e4d17eea)): `validateSqlConnectionUrl` accepts PostgreSQL, MySQL, and MSSQL schemes (rejects MariaDB with guidance), and `validateMongoConnectionUrl` accepts `mongodb` and `mongodb+srv` (requires lowercase scheme)
> - Adds `trimConnectionUrl` filter and `skipWhenConnectionUrl` predicate so blank URL answers fall back to the existing field prompts
> - Extends `option-parser.optionToInquirer` to treat explicitly secret options as masked password prompts and forward configured filters to Inquirer
> - Makes dialect detection case-insensitive in [getDialect](https://github.com/ForestAdmin/toolbelt/pull/807/files#diff-e48af1ebad8eb6d09ed721002096b6f9656989becf4358c0e3bab990e4d17eea) and [Database.getDialect](https://github.com/ForestAdmin/toolbelt/pull/807/files#diff-c94adac0e3963d416860f1fea11bc7a53c8acaf1d222694f31a91a4beb52d7b8) by lowercasing only the scheme comparison while preserving the original URL value
> - Behavioral Change: connection URLs with uppercase or mixed-case schemes now resolve correctly; MariaDB URLs are explicitly rejected in the SQL prompt with a message directing users to MySQL
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 826d38c.</sup>
> 
>

### DIFF
--- a/src/abstract-project-create-command.ts
+++ b/src/abstract-project-create-command.ts
@@ -172,7 +172,7 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
       appPort: Number(config.applicationPort),
     } as AppConfig;
     const dbConfig = {
-      dbConnectionUrl: config.databaseConnectionURL,
+      dbConnectionUrl: config.databaseConnectionURL || undefined, // blank prompt ⇒ use the fields
       dbDialect: config.databaseDialect,
       dbSchema: config.databaseSchema,
       dbName: config.databaseName,
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -14,6 +14,17 @@ export default class NosqlCommand extends AbstractProjectCreateCommand {
     // Ask for a connection URL first; when provided, the field prompts below are skipped.
     databaseConnectionURL: {
       ...projectCreateOptions.databaseConnectionURL,
+      // Mirror the field options' `exclusive: ['databaseConnectionURL']`: when any DB field
+      // flag is passed on the command line, the URL prompt is dropped (scripted/CI usage
+      // must not hang on a new interactive question).
+      exclusive: [
+        'databaseName',
+        'databaseHost',
+        'databasePort',
+        'databaseUser',
+        'databasePassword',
+        'mongoDBSRV',
+      ],
       validate: projectCreateOptions.validateMongoConnectionUrl,
       prompter: {
         question: 'MongoDB connection URL (leave blank to enter the details manually):',
--- a/src/commands/projects/create/nosql.ts
+++ b/src/commands/projects/create/nosql.ts
@@ -11,18 +11,45 @@ import { optionsToFlags } from '../../../utils/option-parser';
 
 export default class NosqlCommand extends AbstractProjectCreateCommand {
   protected static readonly options: CommandOptions = {
-    databaseConnectionURL: projectCreateOptions.databaseConnectionURL,
-    databaseName: projectCreateOptions.databaseName,
-    databaseHost: projectCreateOptions.databaseHost,
-    databasePort: projectCreateOptions.databasePort,
-    databaseUser: projectCreateOptions.databaseUser,
-    databasePassword: projectCreateOptions.databasePassword,
+    // Ask for a connection URL first; when provided, the field prompts below are skipped.
+    databaseConnectionURL: {
+      ...projectCreateOptions.databaseConnectionURL,
+      validate: projectCreateOptions.validateMongoConnectionUrl,
+      prompter: {
+        question: 'MongoDB connection URL (leave blank to enter the details manually):',
+      },
+    },
+    databaseName: {
+      ...projectCreateOptions.databaseName,
+      when: projectCreateOptions.skipWhenConnectionUrl,
+    },
+    databaseHost: {
+      ...projectCreateOptions.databaseHost,
+      when: projectCreateOptions.skipWhenConnectionUrl,
+    },
+    databasePort: {
+      ...projectCreateOptions.databasePort,
+      when: projectCreateOptions.skipWhenConnectionUrl,
+    },
+    databaseUser: {
+      ...projectCreateOptions.databaseUser,
+      when: projectCreateOptions.skipWhenConnectionUrl,
+    },
+    databasePassword: {
+      ...projectCreateOptions.databasePassword,
+      when: projectCreateOptions.skipWhenConnectionUrl,
+    },
 
     // Set prompter to null to replicate bug from previous version (we don't ask for SSL there).
     databaseSslMode: { ...projectCreateOptions.databaseSslMode, prompter: null },
     databaseSSL: { ...projectCreateOptions.databaseSSL, prompter: null },
 
-    mongoDBSRV: projectCreateOptions.mongoDBSRV,
+    mongoDBSRV: {
+      ...projectCreateOptions.mongoDBSRV,
+      when: (args: ProjectCreateOptions) =>
+        projectCreateOptions.skipWhenConnectionUrl(args) &&
+        projectCreateOptions.getDialect(args) === 'mongodb',
+    },
     applicationHost: projectCreateOptions.applicationHost,
     applicationPort: projectCreateOptions.applicationPort,
     language: projectCreateOptions.language,
--- a/src/services/projects/create/options.ts
+++ b/src/services/projects/create/options.ts
@@ -38,6 +38,25 @@ export function getDialect(options: ProjectCreateOptions): ProjectCreateOptions[
   return null;
 }
 
+// Skip a field prompt when the user has provided a connection URL instead (used by sql/nosql).
+export const skipWhenConnectionUrl = (args: ProjectCreateOptions): boolean =>
+  !args.databaseConnectionURL;
+
+// Accept a blank value (⇒ fill the fields instead) or a real `scheme://…` for the engine family.
+export function validateSqlConnectionUrl(value: string): boolean | string {
+  if (!value) return true;
+
+  return /^(postgres|postgresql|mysql|mariadb|mssql):\/\/.+/i.test(value)
+    ? true
+    : 'Enter a URL like postgres://…, mysql://…, mariadb://… or mssql://…';
+}
+
+export function validateMongoConnectionUrl(value: string): boolean | string {
+  if (!value) return true;
+
+  return /^mongodb(\+srv)?:\/\/.+/i.test(value) ? true : 'Enter a mongodb:// or mongodb+srv:// URL';
+}
+
 export const applicationHost: Option = {
   default: 'http://localhost',
   validate: validateAppHostname,
--- a/src/services/projects/create/options.ts
+++ b/src/services/projects/create/options.ts
@@ -43,18 +43,23 @@ export const skipWhenConnectionUrl = (args: ProjectCreateOptions): boolean =>
   !args.databaseConnectionURL;
 
 // Accept a blank value (⇒ fill the fields instead) or a real `scheme://…` for the engine family.
+// The scheme must be lowercase: getDialect() and the generated project both match schemes
+// case-sensitively, so an uppercase scheme would silently produce a project with no dialect.
+// mariadb:// is not offered: the mariadb driver is not shipped, use mysql:// instead.
 export function validateSqlConnectionUrl(value: string): boolean | string {
   if (!value) return true;
 
-  return /^(postgres|postgresql|mysql|mariadb|mssql):\/\/.+/i.test(value)
+  return /^(postgres|postgresql|mysql|mssql):\/\/.+/.test(value)
     ? true
-    : 'Enter a URL like postgres://…, mysql://…, mariadb://… or mssql://…';
+    : 'Enter a URL like postgres://…, mysql://… or mssql://… (the scheme must be lowercase)';
 }
 
 export function validateMongoConnectionUrl(value: string): boolean | string {
   if (!value) return true;
 
-  return /^mongodb(\+srv)?:\/\/.+/i.test(value) ? true : 'Enter a mongodb:// or mongodb+srv:// URL';
+  return /^mongodb(\+srv)?:\/\/.+/.test(value)
+    ? true
+    : 'Enter a mongodb:// or mongodb+srv:// URL (the scheme must be lowercase)';
 }
 
 export const applicationHost: Option = {
--- a/src/commands/projects/create/sql.ts
+++ b/src/commands/projects/create/sql.ts
@@ -12,14 +12,40 @@ import { optionsToFlags } from '../../../utils/option-parser';
 
 export default class SqlCommand extends AbstractProjectCreateCommand {
   protected static options: CommandOptions = {
-    databaseConnectionURL: projectCreateOptions.databaseConnectionURL,
-    databaseDialect: projectCreateOptions.databaseDialectSqlV2,
-    databaseName: projectCreateOptions.databaseName,
+    // Ask for a connection URL first; when provided, the field prompts below are skipped
+    // (dialect/host/port/user/pass/name are derived from the URL).
+    databaseConnectionURL: {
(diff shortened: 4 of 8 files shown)


src/services/projects/create/options.ts
```diff
@@ -52,26 +53,51 @@ export const applicationPort: Option = {
...
+  ],
+  default: 'form',
+  prompter: { question: 'How do you want to connect to your database?' },
+};
+
 export const databaseConnectionURL: Option = {
+  // Prompted only when the user chose the URL method; still usable non-interactively via -c.
+  when: args => args.connectionMethod === 'URI',
+  // Reuse getDialect to reject a URL whose scheme we don't support.
+  validate: value =>
```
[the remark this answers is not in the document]
Gawtier: Fixed in `1e3a9ab` (reworked): the URL is no longer validated via `getDialect`. `create:sql`/`create:nosql` now use `validateSqlConnectionUrl` / `validateMongoConnectionUrl`, which require a real `scheme://…` (regex) and reject `postgres-not-a-url`, `mongodb-invalid`, and cross-engine URLs. Covered by `options.test.ts`.

PMerlet (approved):

src/abstract-project-create-command.ts
```diff
@@ -217,6 +217,11 @@ export default abstract class AbstractProjectCreateCommand extends AbstractAuthe
   protected async getCommandOptions(): Promise<ProjectCreateOptions> {
     const options = await this.optionParser.getCommandLineOptions<ProjectCreateOptions>(this);
 
+    // A blank answer to the optional connection URL prompt means "use the field prompts instead",
+    // and a pasted URL routinely carries surrounding whitespace. Normalize once, before the
+    // dialect is derived from it and before it reaches the connection test and the generated .env.
+    options.databaseConnectionURL = options.databaseConnectionURL?.trim() || undefined;
```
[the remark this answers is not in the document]
PMerlet: Fixed in 9544241 — real bug, thanks.

Root cause is one step earlier than this line: `getCommandLineOptions` only ran an option's `filter` on the interactive answer, never on the flag. So `-c '   '` stayed *defined* through `getInteractiveOptions`, where every database field option is `exclusive` with `databaseConnectionURL` — the dialect/name/host/port/user/password questions were all dropped — and only then did this normalization turn it into `undefined`, leaving `getConfig` with neither a URL nor a dialect.

So the fix normalizes before the skip logic, as suggested, but in the parser rather than in `skipWhenConnectionUrl` (which only covers the prompt path, not the `exclusive` filtering): the flag now goes through the same `filter` inquirer applies to an answer, and a value the filter empties is deleted, so the option is asked for exactly as if the flag had been omitted. Only options declaring a `filter` opt into this, so `--databaseSchema ''` still means "no schema" rather than "ask me".

Covered by `test/utils/option-parser.unit.test.ts` (new): `-c '  '` now asks the URL question and every field question; a padded URL is trimmed and still skips the fields; an empty `--databaseSchema` is untouched. Both URL assertions fail on the previous commit.

src/utils/option-parser.ts
```diff
@@ -88,6 +97,17 @@ export async function getCommandLineOptions<T>(instance: Command): Promise<T> {
...
     if (choice) optionsFromCli[k] = choice.value;
 
+    // Normalize the flag the way inquirer normalizes an answer (it runs `filter` first), so that
+    // both paths agree on what the value is. A value the filter empties was never provided:
+    // dropping it keeps the option out of the interactive skip logic below, instead of silently
+    // suppressing the questions it is exclusive with (`-c '  '` used to skip every database
+    // field prompt and leave the command with neither a connection URL nor a dialect).
+    if (v.filter && typeof optionsFromCli[k] === 'string') {
+      const filtered = v.filter(optionsFromCli[k]);
+      if (filtered === '') delete optionsFromCli[k];
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: a CI run that passes an unset variable to `-c` now hangs on a prompt instead of failing fast.

`option-parser.ts:107` drops the emptied flag, so the connection URL question is asked and inquirer never resolves on a non-terminal stdin.

A way out: keep the delete when stdin is a TTY and reject an empty `-c` otherwise.
</details>

1. Leave `DATABASE_URL` unset in a CI job.
2. Run `forest projects:create:sql app -c "$DATABASE_URL" -H localhost -P 3310 -l javascript`.
3. The empty flag is dropped and the connection URL question is asked.
4. Stdin is not a terminal, so the job hangs until its own timeout.
</details>

src/utils/option-parser.ts
```diff
@@ -88,6 +97,17 @@ export async function getCommandLineOptions<T>(instance: Command): Promise<T> {
...
+      else optionsFromCli[k] = filtered;
+    }
+
     // Validate
     const error = optionsFromCli[k] !== undefined && v.validate?.(optionsFromCli[k]);
     if (typeof error === 'string') throw new Error(`Invalid value for ${k}: ${error}`);
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: a user whose `-c` value is rejected is told to open a GitHub issue about their own typo.

`option-parser.ts:113` throws a plain `Error`, and `abstract-project-create-command.ts:114` prints `ERROR_UNEXPECTED` above it because no `status` is set.

A way out: throw a marked `InvalidOptionError` and print it bare in the catch at `:114`.
</details>

1. Run `forest projects:create:sql app -c 'mariadb://u:p@h:3306/db'`.
2. Log in when asked.
3. Read the GitHub issue request printed above the mariadb hint.
</details>

src/commands/projects/create/nosql.ts
```diff
@@ -11,18 +11,59 @@ import { optionsToFlags } from '../../../utils/option-parser';
...
     // Set prompter to null to replicate bug from previous version (we don't ask for SSL there).
     databaseSslMode: { ...projectCreateOptions.databaseSslMode, prompter: null },
     databaseSSL: { ...projectCreateOptions.databaseSSL, prompter: null },
 
-    mongoDBSRV: projectCreateOptions.mongoDBSRV,
+    mongoDBSRV: {
+      ...projectCreateOptions.mongoDBSRV,
+      when: (args: ProjectCreateOptions) =>
+        projectCreateOptions.skipWhenConnectionUrl(args) &&
+        projectCreateOptions.getDialect(args) === 'mongodb',
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: nobody filling the nosql fields by hand is asked about SRV, so `mongoDBSRV` stays false.

`nosql.ts:65` requires a mongodb dialect that `nosql.ts:99` only sets after prompting returns, so the predicate is never true.

A way out: reduce the predicate to `skipWhenConnectionUrl` and cover the nosql blank-URL branch in `connection-url-prompt.test.ts`.
</details>

1. Run `forest projects:create:nosql app` and leave the connection URL blank.
2. Answer the database field questions.
3. The SRV question never appears and the project is generated for `mongodb://`.
</details>

PMerlet (commented):
**Spec**: no Linear ticket is linked on the pull request, the branch, the title or the commit trailers, so the functional check was limited to the PR description. Against it the change conforms.

**Applies to:** the PR's commit messages
Claude Opus 5 (claude-opus-5): **Should fix**: a script using `-c mariadb://` breaks on a release that semver calls backward compatible.

`.releaserc.js:7` uses the angular preset, where only a `BREAKING CHANGE:` footer forces a major, and no commit here carries one.

A way out: add a `BREAKING CHANGE:` footer naming the rejected schemes to one commit before merging.
</details>

1. Merge this PR into `main`.
2. semantic-release reads the commits and finds no breaking-change footer.
3. A caret install upgrades and `forest projects:create:sql app -c mariadb://…` exits 1.
</details>

src/utils/option-parser.ts
PMerlet: Claude Opus 5 (claude-opus-5): An emptied `-c` is now refused where the message names the flag, instead of being dropped into a question a script cannot answer.

8e896f6

I took the refusal rather than the TTY branch the route suggested, because a flag carrying nothing is a mistake in both modes.
A blank answer to the prompt still falls back to the field prompts, which is what the earlier review asked for.
</details>

src/utils/option-parser.ts
PMerlet: Claude Opus 5 (claude-opus-5): Option validation now throws `InvalidOptionError`, and the create command prints it on its own without the unexpected-error banner.

8e896f6

`InvalidOptionError` extends `ForestCLIError`, the shape this repo already uses for a user-facing failure.
I matched only that class in the catch, so an empty database or a dumper failure keeps the banner it has today.
</details>

src/commands/projects/create/nosql.ts
PMerlet: Claude Opus 5 (claude-opus-5): The three nosql prompts that consulted a dialect set after prompting no longer do, and the blank-URL branch is covered.

8e896f6

Writing the test you asked for showed the same sentence broke two more prompts on that branch.
The port offered no default, so pressing enter was refused by the port validator and the flow could not be completed.
The user was offered the SQL `root`, and both are fixed alongside the SRV predicate.
</details>

PMerlet:
Claude Opus 5 (claude-opus-5): Answering the release finding: the fixes commit carries a `BREAKING CHANGE:` footer naming the rejected schemes, so semantic-release cuts a major.

8e896f6

This repository squash-merges with `COMMIT_MESSAGES`, so the footer reaches the squashed body the commit analyzer reads.
Read off the repository settings rather than assumed.
</details>

PMerlet:
Claude Opus 5 (claude-opus-5): the branch was rewritten to drop a `BREAKING CHANGE:` footer that no longer applied, so the sha my earlier replies name has moved.

| then | now |
| --- | --- |
| `8e896f6` | `afb7c0d` |
| `e0d43a6` | `75d494d` |

The connection URL validators no longer gate the `-c` flag, so nothing a script could run before stops working and the footer was false.
`git diff` between the old tip and the new one is empty, so only the messages and the shas changed.
`9544241` and every commit before it are untouched.
</details>

PR #816, opened 2026-08-20
feat(services): add a process runner that can actually stop what it started

Gawtier:
Groundwork for `forest start` (#next), which has to drive `npm install`, `npm start`, `bundle add` and `bin/rails server` — the toolbelt has no way to run a long-lived process today.

## Why a service, and why it looks like this

One property of package managers explains the whole design: **`npm start` is a wrapper**. The process holding the port is its *child*, not the one we spawned. Signalling what we spawned leaves the server running **and** keeps this CLI alive on its still-open pipes — which is exactly how "Stop" ended up doing nothing in the onboarding prototype.

So long-running processes are started in their own process group, and stopped by signalling that group. And stopping means stopped: SIGTERM is a request a back-end is free to trap and decline, so the group is killed outright if it is still there after a grace period.

## What it provides

| | |
| :-- | :-- |
| `runStep` | inherits stdio, so `forest login` can open a browser and a package manager's prompts render natively instead of being buffered into silence |
| `runCapture` | captures stdout and stderr **separately**, streams stderr through `onProgress`, and puts the sub-command's own message in the error |
| `startProcess` | streams output, resolves once the process announces itself, fails fast on a taken port, returns `exited` and `mute()` |
| `stopProcess` | signals the group, escalates to SIGKILL, reads the exit state, never throws |

Credentials never reach an error message: passwords are stripped from anything URL-shaped in both the arguments and the captured output, and the value of a flag whose name says it is a secret is dropped whole. The host and database survive, because they are what makes a failure diagnosable and they are not the secret.

## Findings folded in

Two adversarial reviews ran against this branch. Every defect below was reproduced first, and each is covered by a test that fails without its fix.

**First round — on the original implementation:**

- **Listeners were never released** once ready settled, so the buffer grew for the process's whole life and the ready regex re-ran over all of it on every chunk — quadratic, measured at 33 MB of heap after 8 s of logs, on a service designed to run for hours.
- **A ready rejection left the child running**, and its open pipes kept the CLI's event loop alive: "Timed out…" followed by a prompt that never came back.
- **`detached: true` meant Ctrl-C never reached the child**, so any interrupt or crash stranded a server holding a port. Started processes are tracked and stopped on exit and on SIGINT/SIGTERM/SIGHUP.
- **`stopProcess` guarded on `child.killed`**, which `process.kill()` never sets — so it re-signalled a pid the OS may have reused.
- **`runCapture` merged the streams**, which breaks by construction against a command whose stdout is a machine-readable document and whose progress goes to stderr.

**Second round — five more, on the hardened version:**

- **`stopProcess` did nothing at all once the wrapper had exited.** It read the *leader's* exit state, but the wrapper can return while the server it launched keeps the port — the exact failure this module exists to prevent, reached from the other side. What must not be repeated is the signal, so that is what is tracked now; a signal-0 probe on the group rules out the recycled pid the old guard was there for, and POSIX forbids recycling a pid while a process group still carries it as its id.
- **Nothing escalated past SIGTERM**, so a server that traps it — puma, and most back-ends — kept the port for good. The group is now killed outright after a grace period, on an unref'd timer. On the way out that timer can never run, so the exit hook takes the same grace synchronously.
- **Every argument was pasted into the failure message**, with the captured stderr on top, so a connection URL or a registry token was handed straight back into the terminal and into whatever collects its output. The CLI masks that connection URL at the prompt; it must not print it back in the failure it caused.
- **EADDRINUSE was fatal even when the process recovered from it.** Plenty of dev servers report the clash and then bind the next port up; an unanchored substring match killed a back-end that was about to serve. It starts a countdown now, which announcing readiness cancels — and the countdown does not depend on finding the port in the message, which Ruby writes before the word and a unix socket does not write at all.
- **A crash after `ready` was silent.** `ready` only ever describes the start, so a back-end that died twenty minutes in left a caller with no way to notice. `exited` resolves whenever the process ends, carrying the signal as well as the code so a caller can tell a crash from its own stop.

**Third round — Macroscope, on the reviewed version:** a group dropped from the register as soon as its wrapper closed, and again at signal time, putting a still-running server out of reach of the exit hook; "already signalled" remembered instead of asked, so a `SIGKILL` after a declined `SIGTERM` did nothing; a POSIX `child.kill` fallback that could reach a recycled pid; and two credential shapes the redaction was written past — `redis://:secret@host`, and a secret value starting with `-`.

One of its findings was declined: swapping `close` for `exit` in the startup watch would reject a start that succeeds, since a launcher that hands over and leaves exits long before the server it left behind announces itself. That case is now a test.

Also fixed along the way: SIGHUP exited 143, which belongs to SIGTERM — the convention is 128 + the signal number, so 129.

## Tested

**38 tests**, on real processes rather than mocks. The key one spawns a real wrapper whose *child* holds a port, because `sh -c 'cmd'` alone execs into a single process and cannot reproduce the bug; it asserts the port is free after `stopProcess`.

Also covered: the wrapper that exits before its child, the server that traps SIGTERM, the exit hook asserted on the signals it has already sent by the time it calls `process.exit` (the kernel freeing the socket is not synchronous, so asserting on the port there is a race that loses on Linux), the buffer released at ready, the port freed after a failed start, no re-signal after exit, muting, stderr not corrupting a JSON stdout, and credentials absent from every error the module raises.

Ports are asked of the OS rather than hardcoded in the newer tests: these are tests about servers that outlive what started them, so a run that leaves one behind must not be able to poison the next.

⚠️ **Windows is a documented limitation, not an oversight**: `process.kill(-pid)` does not exist there and `detached` creates no signalable group, so the fallback signals the wrapper alone. CI is Linux-only and the onboarding is not offered on Windows; if that changes this needs `taskkill /T /F`.

## Known, and deliberately left

- `installExitHook()` is a side effect of `startProcess`: it installs global signal handlers that are never removed and calls `process.exit()` directly, where the repo has `src/context/process-plan.js` to make that injectable. Changing it means changing who calls what, which is a design decision rather than a fix.
- `runCapture` buffers stdout and stderr without a bound, where `startProcess` is carefully capped — defensible, since the point is to get the whole document.
- Redaction is deliberately narrow: a secret that is neither shaped like a URL nor the value of a secret-looking flag still goes through.
- **Nothing imports this module yet.** That is what `forest start` is for, and it is where the API will first meet its real usage.

> [!NOTE]
> ### Add `process-runner` service with group shutdown and readiness tracking
> - Introduces [process-runner.ts](https://github.com/ForestAdmin/toolbelt/pull/816/files#diff-8792b9366db21ee8c5b5272531b5b3a1e99429c29cd902fc1e0f272d67a98429) to run commands via `runStep`, `runCapture`, and `startProcess`.
> - `startProcess` launches detached background processes, monitors output for readiness or `EADDRINUSE` port clashes, and supports log muting.
> - `stopProcess` and `stopAllProcesses` use process-group signaling on POSIX to stop descendants, escalating from graceful termination to `SIGKILL` after a grace period.
> - Exit hooks clean up tracked process groups synchronously on normal and signal-based exits. Error messages redact secrets in connection URLs and arguments.
> - Risk: `signalGroup` relies on POSIX process groups; Windows falls back to direct child signaling and may leave descendants running.
>
> 
> #### Changes since #816 opened
>
> - Replaced the `stopped` WeakSet with an `escalating` WeakSet in the process runner module to track which `ChildProcess` instances already have an escalation countdown scheduled, allowing `stopProcess` to send additional signals to still-alive process groups while preventing multiple escalation timers from being stacked on the same child, and modified the `close` event handler in `startProcess` to only remove a child from the `running` registry when `groupAlive` confirms the entire process group has terminated rather than when the wrapper process exits [3c8b07e]
> - Modified the `signalGroup` utility to explicitly check `CAN_SIGNAL_GROUPS` and attempt `child.kill(signal)` on platforms without group signaling support before returning, and on POSIX-capable systems to signal the negative pid for the process group while ignoring failures as already dead, removing the previous implicit fallback from group signaling failure to signaling the leader pid [3c8b07e]
> - Updated the `stopProcess` function to change its early-return guard to only check for an existing pid and whether the process group is alive via `groupAlive`, removing reliance on the `stopped` WeakSet and no longer deleting the child from the `running` registry within the function, and to always attempt signaling the group via `signalGroup` and allow `SIGKILL` to be sent immediately even if a prior `SIGTERM` was sent [3c8b07e]
> - Adjusted the `URL_CREDENTIALS` regular expression to allow an empty username segment before the password by changing the quantifier from `+` to `*`, enabling redaction of credentials in URLs with patterns like `redis://:password@host`, and introduced a `LONG_FLAG` constant to detect double-dash flags and updated `redactArgs` to consider a secret value as the next token unless it matches `LONG_FLAG` rather than any token starting with a dash [3c8b07e]
> - Added test cases covering processes remaining reachable for exit hooks when trapping signals, servers remaining in the running registry after their wrapper processes exit, waiting for readiness even when wrappers exit early, redacting credentials in URLs with empty usernames, and redacting secret values beginning with a single dash without consuming subsequent long flags, and updated existing tests to reflect the new behavior where `stopProcess` does not suppress re-signaling and probes group liveness via signal 0 [3c8b07e]
> - Modified `startProcess` function to wait for entire process group termination before resolving the `exited` promise, introduced `ended` WeakSet to permanently track terminated process groups, and updated `groupAlive` function to mark groups as ended and remove them from the `running` set when detected as not alive [e1302de]
> - Modified `watchStartup` function buffering logic to concatenate prior window with new chunks before classification to detect readiness tokens split across I/O boundaries [e1302de]
> - Introduced `REAP_POLL_MS` constant set to 250 milliseconds defining polling frequency for process group liveness checks after wrapper process exits [e1302de]
> - Modified `readPortClash` utility to strip time-like tokens from log lines before extracting port numbers to avoid misinterpreting timestamp components as ports [e1302de]
> - Extracted `probeGroup` function to centralize process group liveness checking logic using signal 0 on the negative pid [e1302de]
> - Updated and expanded test coverage for process runner to verify process group lifecycle management, readiness detection across chunk boundaries, timestamp handling in port clash detection, and that ended groups are not probed or signaled after termination [e1302de]
> - Replaced single-buffer stream output detection with sliding window scanning to detect ready patterns and port clashes that span read boundaries [7d8fc75]
> - Encapsulated process group end tracking into reusable helper [7d8fc75]
> - Consolidated POSIX and Windows process signaling paths in `signalGroup` util [7d8fc75]
> - Introduced synchronous blocking helper and updated synchronous shutdown to use it [7d8fc75]
> - Refactored port clash detection in `readPortClash` util with reusable time-stripping regex [7d8fc75]
> - Added stream scanning and time pattern constants [7d8fc75]
> - Updated comments across implementation and test files for clarity [7d8fc75]
> - Replaced `escalating` WeakSet with `escalations` WeakMap in process-runner to store escalation timeout handles per child process, and modified `stopProcess` to clear any existing escalation timer before scheduling a new one [af0a821]
> - Replaced regex-based secret flag detection with word-boundary-aware detection using `isSecretFlag` helper function [af0a821]
> - Updated test suite to use dynamically allocated ports instead of fixed ports, reduced timeout values in EADDRINUSE tests, and added tests verifying escalation timer replacement and word-boundary-based flag redaction [af0a821]
> - Implemented suffix-based secret detection for command flags with digit stripping and exclusion list [2b92710]
> - Modified port clash detection to prioritize clashes containing port values [2b92710]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 761411b.</sup>
> 
>

### DIFF
--- a/src/services/process-runner.ts
+++ b/src/services/process-runner.ts
@@ -140,11 +140,15 @@ export function runCapture(
     let stdout = '';
     let stderr = '';
 
-    child.stdout?.on('data', chunk => {
-      stdout += chunk.toString();
+    // Decoded by the stream, not per chunk: a multibyte character split across two reads would
+    // otherwise become two replacement characters — silently changing a value we then JSON.parse.
+    child.stdout?.setEncoding('utf8');
+    child.stderr?.setEncoding('utf8');
+
+    child.stdout?.on('data', (text: string) => {
+      stdout += text;
     });
-    child.stderr?.on('data', chunk => {
-      const text = chunk.toString();
+    child.stderr?.on('data', (text: string) => {
       stderr += text;
       onProgress?.(text);
     });
@@ -228,8 +232,7 @@ export function startProcess(
       reject(error);
     };
 
-    const onData = (chunk: Buffer) => {
-      const text = chunk.toString();
+    const onData = (text: string) => {
       stream?.(text);
       if (settled) return;
       scanned += text;
@@ -247,6 +250,10 @@ export function startProcess(
       timeoutMs,
     );
 
+    // Same reason as `runCapture`, with a sharper consequence: a `ready` pattern straddling a
+    // chunk boundary would never match, and the process would be killed on a false timeout.
+    child.stdout?.setEncoding('utf8');
+    child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', onData);
     child.stderr?.on('data', onData);
     child.on('error', error => fail(error));
--- a/src/services/process-runner.ts
+++ b/src/services/process-runner.ts
@@ -219,17 +219,26 @@ export function startProcess(
   };
 
   const readyPromise = new Promise<void>((resolve, reject) => {
-    // Only accumulated until `ready` matches, then released. Keeping it would mean re-running the
-    // regex over an ever-growing string for the process's whole life — quadratic, on a buffer a
-    // long-lived server grows to hundreds of megabytes.
-    let scanned = '';
+    // Rebuilt without `g`/`y`: those flags make `.test()` stateful, so a caller passing `/x/g`
+    // would have its `lastIndex` advance between chunks and skip the very announcement we wait
+    // for — the process then dies on a timeout that had no cause.
+    const readyPattern = new RegExp(ready.source, ready.flags.replace(/[gy]/g, ''));
+
+    // One buffer PER STREAM. Sharing one would let a token straddling stdout and stderr match
+    // when neither stream ever produced it. Each is capped to a suffix: a chatty process that
+    // never announces itself would otherwise exhaust the heap long before the timeout fires,
+    // crashing instead of reporting. The window is far wider than any readiness line, so a
+    // pattern split across chunk boundaries still matches.
+    const WINDOW = 8192;
+    const scanned = { stdout: '', stderr: '' };
     let settled = false;
     let timeout: NodeJS.Timeout;
 
     const settle = () => {
       settled = true;
       clearTimeout(timeout);
-      scanned = ''; // release the buffer; `onData` keeps streaming but stops scanning
+      scanned.stdout = '';
+      scanned.stderr = '';
     };
 
     // A failed start must not leave the process behind: its open pipes would also keep this CLI's
@@ -241,15 +250,15 @@ export function startProcess(
       reject(error);
     };
 
-    const onData = (text: string) => {
+    const onData = (source: 'stdout' | 'stderr') => (text: string) => {
       stream?.(text);
       if (settled) return;
-      scanned += text;
+      scanned[source] = (scanned[source] + text).slice(-WINDOW);
 
-      if (ready.test(scanned)) {
+      if (readyPattern.test(scanned[source])) {
         settle();
         resolve();
-      } else if (/EADDRINUSE/.test(scanned)) {
+      } else if (/EADDRINUSE/.test(scanned[source])) {
         fail(new Error('Port already in use — free it with `lsof -ti :<port> | xargs kill`.'));
       }
     };
@@ -259,12 +268,13 @@ export function startProcess(
       timeoutMs,
     );
 
-    // Same reason as `runCapture`, with a sharper consequence: a `ready` pattern straddling a
-    // chunk boundary would never match, and the process would be killed on a false timeout.
+    // Decoded by the stream, not per chunk: a multibyte character split across two reads would
+    // otherwise become replacement characters, and a `ready` pattern containing one would never
+    // match — the process killed on a false timeout.
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
-    child.stdout?.on('data', onData);
-    child.stderr?.on('data', onData);
+    child.stdout?.on('data', onData('stdout'));
+    child.stderr?.on('data', onData('stderr'));
     child.on('error', error => fail(error));
     child.on('close', code =>
       fail(new Error(`\`${command}\` stopped before it was ready (exit code ${code}).`)),
--- a/src/services/process-runner.ts
+++ b/src/services/process-runner.ts
@@ -11,6 +11,11 @@ import { spawn } from 'child_process';
  * signals the wrapper, leaves the server running, and — since its pipes stay open — keeps this
  * CLI alive too. Every long-running process is therefore started in its own process group, and
  * stopped by signalling that group.
+ *
+ * KNOWN LIMITATION — Windows. `process.kill(-pid)` does not exist there and `detached` creates no
+ * signalable group, so `stopProcess` falls back to signalling the wrapper alone: the server it
+ * spawned survives, which is the very bug this module fixes elsewhere. CI is Linux-only and the
+ * onboarding is not offered on Windows; if that changes, this needs `taskkill /T /F`.
  */
 
 export type RunOptions = {
@@ -22,9 +27,65 @@ export type StartedProcess = {
   child: ChildProcess;
   /** Resolves when the process prints something matching `ready`, rejects on timeout or a taken port. */
   ready: Promise<void>;
+  /** Stop streaming output — before handing the terminal to something else, typically. */
+  mute: () => void;
 };
 
+export type CaptureResult = { stdout: string; stderr: string };
+
 const READY_TIMEOUT_MS = 120_000;
+
+/**
+ * Every process we started and have not stopped. Registered so the CLI can take them down with it:
+ * a detached child survives its parent by design, and the terminal's Ctrl-C never reaches it (it
+ * sits in its own process group), so without this a crash or an interrupt strands a server holding
+ * a port the user then has to hunt down with `lsof`.
+ */
+const running = new Set<ChildProcess>();
+let exitHookInstalled = false;
+
+/**
+ * Stop a process started by `startProcess`, and everything it spawned.
+ *
+ * A negative pid signals the whole process group — the only way to reach the server a package
+ * manager launched on our behalf. Never throws: stopping something already stopped is a success.
+ */
+export function stopProcess(child: ChildProcess | undefined, signal: NodeJS.Signals = 'SIGTERM') {
+  // `child.killed` is not the state we need: it only records that `child.kill()` was called, and
+  // the group path uses `process.kill()`, which never sets it. `exitCode`/`signalCode` are what
+  // actually say the process is gone — and they stop us re-signalling a pid the OS may have reused.
+  if (!child?.pid || child.exitCode !== null || child.signalCode !== null) return;
+
+  running.delete(child);
+
+  try {
+    process.kill(-child.pid, signal);
+  } catch {
+    try {
+      child.kill(signal);
+    } catch {
+      // Already dead — nothing left to stop.
+    }
+  }
+}
+
+function installExitHook() {
+  if (exitHookInstalled) return;
+  exitHookInstalled = true;
+
+  const stopAll = () => running.forEach(child => stopProcess(child));
+
+  // `exit` covers a normal end and an uncaught throw. The signals cover the terminal, which would
+  // otherwise kill this process and leave the group behind. `once` so a second Ctrl-C is never
+  // swallowed — the user must always be able to give up.
+  process.on('exit', stopAll);
+  (['SIGINT', 'SIGTERM', 'SIGHUP'] as const).forEach(signal =>
+    process.once(signal, () => {
+      stopAll();
+      process.exit(signal === 'SIGINT' ? 130 : 143);
+    }),
+  );
+}
 
 function spawnOptions(options: RunOptions, extra: SpawnOptions = {}): SpawnOptions {
   return {
@@ -34,12 +95,12 @@ function spawnOptions(options: RunOptions, extra: SpawnOptions = {}): SpawnOptio
   };
 }
 
+const formatCommand = (command: string, args: string[]) => `${command} ${args.join(' ')}`.trim();
+
 /**
  * Run a command to completion. stdio is inherited so the child owns the terminal: `forest login`
  * can open a browser, and a package manager's own prompts and progress render natively instead of
  * being buffered into silence.
- *
- * Rejects with the exit code when the command fails — the caller decides whether that is fatal.
  */
 export function runStep(command: string, args: string[], options: RunOptions = {}): Promise<void> {
   return new Promise((resolve, reject) => {
@@ -49,37 +110,61 @@ export function runStep(command: string, args: string[], options: RunOptions = {
     child.on('close', code =>
       code === 0
         ? resolve()
-        : reject(new Error(`\`${command} ${args.join(' ')}\` exited with code ${code}`)),
+        : reject(new Error(`\`${formatCommand(command, args)}\` exited with code ${code}`)),
     );
   });
 }
 
-/** Same, but capturing stdout instead of inheriting it — for commands we need to read back. */
+/**
+ * Run a command, capturing its streams SEPARATELY.
+ *
+ * Keeping them apart is the point: a command whose stdout is a machine-readable document writes
+ * its progress to stderr, so merging the two corrupts the document — the parse then fails silently
+ * and the caller proceeds with nothing. stderr is streamed through `onProgress` instead, so the
+ * user still sees what is happening.
+ *
+ * On failure the error carries the captured stderr: piping it means the sub-command's own message
+ * never reached the terminal, and "exited with code 2" alone tells the user nothing.
+ */
 export function runCapture(
   command: string,
   args: string[],
-  options: RunOptions = {},
-): Promise<string> {
+  { onProgress, ...options }: RunOptions & { onProgress?: (chunk: string) => void } = {},
+): Promise<CaptureResult> {
   return new Promise((resolve, reject) => {
     const child = spawn(
       command,
       args,
       spawnOptions(options, { stdio: ['inherit', 'pipe', 'pipe'] }),
     );
-    let output = '';
+    let stdout = '';
+    let stderr = '';
 
     child.stdout?.on('data', chunk => {
-      output += chunk.toString();
+      stdout += chunk.toString();
     });
     child.stderr?.on('data', chunk => {
-      output += chunk.toString();
+      const text = chunk.toString();
+      stderr += text;
+      onProgress?.(text);
     });
     child.on('error', reject);
-    child.on('close', code =>
-      code === 0
-        ? resolve(output)
-        : reject(new Error(`\`${command} ${args.join(' ')}\` exited with code ${code}`)),
-    );
+    child.on('close', code => {
+      if (code === 0) {
+        resolve({ stdout, stderr });
+
+        return;
+      }
+
+      const detail = (stderr || stdout).trim();
+      reject(
+        new Error(
+          `\`${formatCommand(command, args)}\` exited with code ${code}${
+            detail ? `:\n${detail}` : ''
+          }`,
+        ),
+      );
+    });
   });
 }
 
@@ -88,8 +173,7 @@ export function runCapture(
  * resolve `ready` once it prints something matching `ready` — for an agent, the line that says its
  * schema reached Forest.
  *
- * `detached` is the important part: it makes the child a process-group leader so `stopAgent` can
- * take the whole tree down. Without it, a package-manager wrapper survives its own `kill`.
+ * `detached` makes the child a process-group leader so `stopProcess` can take the whole tree down.
  */
 export function startProcess(
   command: string,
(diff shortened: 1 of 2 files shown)


src/services/process-runner.ts
```diff
@@ -0,0 +1,406 @@
...
+function signalGroup(child: ChildProcess, signal: NodeJS.Signals) {
+  try {
+    process.kill(-(child.pid as number), signal);
+  } catch {
+    try {
+      child.kill(signal);
+    } catch {
+      // Already dead — nothing left to stop.
+    }
+  }
```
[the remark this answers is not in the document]
PMerlet: Taken, in 51f70f1 — with the reasoning pushed a little further. The fallback is now the Windows path only, and not merely restricted to it: on POSIX the leader was *in* the group we just failed to signal, so its pid adds nothing even when it is still valid.

No test — reaching that branch means `kill(-pid)` failing while `groupAlive` has just said the group is there, which I could not provoke without stubbing the syscall.

src/services/process-runner.ts
```diff
@@ -0,0 +1,406 @@
...
+    cwd: options.cwd,
+    env: options.env ? { ...process.env, ...options.env } : process.env,
+    ...extra,
+  };
+}
+
+const SECRET_FLAG =
+  /^--?[a-z0-9-]*(token|secret|password|passwd|pwd|apikey|api-key|auth|credential)[a-z0-9-]*$/i;
+
+const URL_CREDENTIALS = /([a-z][a-z0-9+.-]*:\/\/[^\s/@:]+):[^\s/@]+@/gi;
```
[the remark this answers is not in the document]
PMerlet: Taken, in 3c8b07e. `redis://:password@host` is not an edge case, it is how Redis and Mongo URLs are normally written — which is exactly the shape a `--connection-url` for either of them arrives in. Covered by a test that fails with the `+`.

src/services/process-runner.ts
```diff
@@ -0,0 +1,454 @@
...
+ */
+function redactSecrets(text: string): string {
+  return text.replace(URL_CREDENTIALS, '$1:***@');
+}
+
+function redactArgs(args: string[]): string[] {
+  let valueIsSecret = false;
+
+  return args.map(arg => {
+    const isSecretValue = valueIsSecret && !arg.startsWith('-');
```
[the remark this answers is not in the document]
PMerlet: Real leak, taken in 3c8b07e — but not with this diff, which trades it for a different wrong message.

Redacting unconditionally turns `--auth-token --verbose` into `--auth-token ***`: the error then claims a credential was passed where none was, and drops the flag that was. The point of putting the command in the message is to say what ran.

So only an unmistakable long flag (`/^--[a-z0-9][a-z0-9-]*$/`) is exempt. `-dashy-looking-secret` is redacted, `--verbose` still reads as itself. Both cases are asserted.

src/services/process-runner.ts
```diff
@@ -0,0 +1,468 @@
...
+  installExitHook();
+
+  const child = spawn(
+    command,
+    args,
+    spawnOptions(options, { stdio: ['ignore', 'pipe', 'pipe'], detached: true }),
+  );
+
+  running.add(child);
+  child.on('close', () => running.delete(child));
```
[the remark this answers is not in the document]
PMerlet: Taken, in 51f70f1, and the same mistake was in two more places — thank you, this one pulled the thread.

`stopProcess` also dropped the child at signal time, so a process that traps SIGTERM and outlives a CLI exiting before the unref'd escalation fires was out of reach too. And the `stopped` guard meant `stopAllProcesses('SIGKILL')` after a declined SIGTERM silently did nothing. A child now stays registered until its whole group has ended, and what is tracked is the countdown rather than the fact of having signalled.

src/services/process-runner.ts
```diff
@@ -0,0 +1,508 @@
...
+
+    // Decoded by the stream, not per chunk: a multibyte character split across two reads would
+    // otherwise become replacement characters, and a `ready` pattern containing one would never
+    // match — the process killed on a false timeout.
+    child.stdout?.setEncoding('utf8');
+    child.stderr?.setEncoding('utf8');
+    child.stdout?.on('data', onData('stdout'));
+    child.stderr?.on('data', onData('stderr'));
+    child.on('error', error => fail(error));
+    child.on('close', code =>
```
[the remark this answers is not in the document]
PMerlet: Not taking this one — it would reject a start that succeeds, and I have added a test that fails with the suggested diff.

`sh -c 'server &'` with no `wait`: the launcher is gone in milliseconds and the server it left behind holds the pipes. Measured on this branch:

```
outcome: RESOLVED | events seen before it settled: [ 'exit@657' ]
```

The wrapper's `exit` fires well before the readiness line, and `close` never fires at all. Failing on `exit` would reject that start — which is this module's own thesis, that the wrapper is not the server.

The symptom you describe is real, but rejecting is not the right answer to it: while a descendant is alive the start may still succeed, so the readiness timeout is what should end it. And when the group really is gone, every write end of those pipes is closed with it, so `close` does fire — `exit` plus a liveness check would buy nothing over `close`.

The case is now covered in `a wrapper that exits before the server it started announces itself`, so this does not get changed by accident later.

src/services/process-runner.ts
```diff
@@ -0,0 +1,531 @@
...
+  graceMs: number = STOP_GRACE_MS,
+) {
+  // NOT the leader's exit state. The process we spawned is a wrapper, and it can exit while the
+  // server it launched keeps the port — the very failure this module exists to prevent, reached
+  // from the other side. Whether the group is still there is the question, and `groupAlive` both
+  // answers it and rules out the pid the OS could have recycled. (`child.killed` is no use
+  // either: it only records that `child.kill()` was called, and the group path never calls it.)
+  if (!child?.pid || !groupAlive(child)) return;
+
+  signalGroup(child, signal);
```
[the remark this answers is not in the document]
PMerlet: Taken, in 8fff35d — the conclusion is right, though the way in is not the one described.

A recycled pid was already ruled out while the group is alive: POSIX forbids reusing the number a process group carries as its id. What was missing is the case where the group is *not* alive and we have not noticed — and that case was reachable, because the handle could never leave the register.

`close` waits on the pipes, not on the group. A server that logs to a file closes them with its wrapper while it keeps the port, so `close` fired with `groupAlive` still true and the only `running.delete` in the module never ran again. The handle then stayed registered for the life of the CLI, long past the group's actual end — and from there `-pid` is a number the OS is free to have handed to somebody else. The next Ctrl-C SIGTERMs, then SIGKILLs, a stranger's process group.

What makes this fixable is that the probe is only trustworthy in one direction: "gone" cannot be wrong about us, "still there" can be, once we have stopped being sure. So the first "gone" is recorded as final, and the pid is afterwards neither signalled nor asked about. And since the end nobody is present for is exactly the one that must not be missed, the leader being reaped now starts watching for it — an unref'd poll — instead of concluding anything.

Reproduced first: `stopAllProcesses()` on a group that had ended minutes earlier still probed `-pid`. The test asserts that number is never passed to `process.kill` again, not even with signal 0.

One existing assertion had to change with it. It required that a second `stopProcess` *ask* rather than remember — written against the old `stopped` guard, which remembered having signalled, which says nothing about whether anything stopped. Remembering an end that was observed is a different claim, and the safe one; the test now says so.

src/services/process-runner.ts
```diff
@@ -0,0 +1,603 @@
...
+    };
+
+    child.on('close', resolveIfGone);
+    child.on('exit', () => {
+      if (resolveIfGone()) return;
+
+      const poll = setInterval(() => {
+        if (resolveIfGone()) clearInterval(poll);
+      }, REAP_POLL_MS);
+      poll.unref();
```
PMerlet: Claude Opus 5.5 (claude-opus-5-5): **Should fix**: a command awaiting `exited` on a server that logs to a file exits with code 0 and takes that server down.

`process-runner.ts:149` unrefs the only timer left to resolve `exited`, so Node exits 0 and the exit hook kills the group.

A way out: keep the poll referenced, since `resolveIfGone` already clears it, or say on `exited` that awaiting it holds nothing.
</details>

1. Outside Jest, call `startProcess('sh', ['-c', 'sleep 3 >/dev/null 2>&1 </dev/null & echo ready'], { ready: /ready/ })`.
2. Await `ready`, then await `exited`.
3. Node exits with code 0 after about 15 ms, and `exited` never settles.
4. The `sleep` is killed by the exit hook.
</details>

src/services/process-runner.ts
```diff
@@ -0,0 +1,603 @@
...
+  'password',
+  'pwd',
+  'secret',
+  'token',
+]);
+
+/** OAuth is the protocol, not the credential: `--oauth-token` is one, `--oauth-callback` is not. */
+const NOT_SECRET_WORDS = new Set(['oauth']);
+
+const FLAG_NAME = /^--?([a-z0-9][a-z0-9-]*)$/i;
```
PMerlet: Claude Opus 5.5 (claude-opus-5-5): **Should fix**: a failing command given `--client_secret=…` or `--api_key <value>` prints that secret in its error message.

`FLAG_NAME` at `process-runner.ts:247` and `LONG_FLAG` at line 311 admit no underscore, so a snake_case secret flag goes unrecognised.

A way out: allow `_` in both patterns and split the words on `[-_]`.
</details>

1. Call `runCapture('sh', ['-c', 'exit 1', 'x', '--client_secret=s3cr3t', '--api_key', 'k3y', '--client-secret=ok'])`.
2. Read the rejection message.
3. It shows `s3cr3t` and `k3y` in clear, while `--client-secret=***` is redacted.
</details>

src/services/process-runner.ts
```diff
@@ -0,0 +1,603 @@
...
+      timeoutMs,
+    );
+
+    // Decoded by the stream, not per chunk: a multibyte character split across two reads would
+    // otherwise become replacement characters.
+    child.stdout?.setEncoding('utf8');
+    child.stderr?.setEncoding('utf8');
+    child.stdout?.on('data', onData('stdout'));
+    child.stderr?.on('data', onData('stderr'));
+    child.on('error', error => fail(error));
```
PMerlet: Claude Opus 5.5 (claude-opus-5-5): **Violates conventions**: a regression in how `startProcess` reports a command it cannot spawn would pass the suite unnoticed.

`process-runner.ts:548` rejects `ready` on a spawn `error`, and no case reaches it, against `testing.md#cover-error-and-edge-paths-not-only-the-happy-path`.

A way out: a `startProcess('definitely-not-a-command', …)` case asserting `ready` rejects, as `runStep` does at test line 86.
</details>

src/services/process-runner.ts
```diff
@@ -0,0 +1,603 @@
...
+    });
+    child.on('error', reject);
+    child.on('close', code => {
+      if (code === 0) {
+        resolve({ stdout, stderr });
+
+        return;
+      }
+
+      const detail = redactSecrets((stderr || stdout).trim());
```
PMerlet: Claude Opus 5.5 (claude-opus-5-5): **Violates conventions**: a failing command that explains itself only on stdout has no test for its message or its redaction.

`process-runner.ts:389` falls back to stdout, and no failing case writes only there, against `testing.md#cover-error-and-edge-paths-not-only-the-happy-path`.

A way out: a `runCapture` case that prints `postgres://u:pw@host` with `console.log`, exits non-zero, and asserts `***` in the message.
</details>

PMerlet (commented):
**Spec**: no Linear ticket is linked, so the functional check was limited to the PR description.

src/services/process-runner.ts
PMerlet: Claude Opus 5.5 (claude-opus-5-5): The reap poll in `trackGroup` is no longer unref'd, so awaiting `exited` now holds the CLI open until the group ends.
4f25ce9

Keeping it referenced costs nothing once the group is gone, because `resolveIfGone` clears it on the tick the group ends.
The new test asserts both halves: the poll holds the loop while waiting, and `clearInterval` receives it afterwards.
</details>

src/services/process-runner.ts
PMerlet: Claude Opus 5.5 (claude-opus-5-5): Names are now split on `-` and `_`, and a `NAME=value` argument is read the same way, so all three shapes are redacted.
4f25ce9

The defect was any secret-named argument outside kebab or camel case, not only the underscore flag.
`FOREST_ENV_SECRET=…` passed to `env` or `docker run -e` had the same gap, and `NODE_ENV=production` still reads as itself.
</details>

src/services/process-runner.ts
PMerlet: Claude Opus 5.5 (claude-opus-5-5): Added a `startProcess` case where the command does not exist, asserting that `ready` rejects with `ENOENT`.
4f25ce9

src/services/process-runner.ts
PMerlet: Claude Opus 5.5 (claude-opus-5-5): Added a `runCapture` case failing on stdout alone, asserting its message is carried and its connection URL is redacted.
4f25ce9

src/services/process-runner.ts
PMerlet: Closed: awaiting `exited` now holds the CLI until the whole group ends, because the reap poll stays referenced.
<details><summary>How it was checked</summary>

Predicate: the leader is reaped, the pipes are closed, the group is alive, and nothing holds the event loop.
Executed at 4f25ce9 outside Jest, `exited` settled after 3 s, and a stopped SIGTERM-trapping server still escalated and let the CLI exit.
Pinned: `keeps the CLI alive while it waits on a group whose pipes have closed` fails on the pre-fix code.
</details>

src/services/process-runner.ts
PMerlet: Closed: a flag or `NAME=value` argument whose name says it is secret is now redacted in snake, kebab and camel case alike.
<details><summary>How it was checked</summary>

Predicate: an argument whose name's words include a secret word reaches the message with its value intact.
Executed at 4f25ce9, nine secret shapes came out as `***` while `--no_token`, `NODE_ENV=` and `--oauth_callback=` read as themselves.
Pinned: `reads a snake_case flag and a NAME=value assignment as words too` fails on the pre-fix code.
</details>

src/services/process-runner.ts
PMerlet: Closed: a `startProcess` case now reaches the spawn-error branch, as `testing.md#cover-error-and-edge-paths-not-only-the-happy-path` asks.
<details><summary>How it was checked</summary>

The new case asserts `ready` rejects with `ENOENT`, which only the `error` handler at line 560 produces.
Ran it at 4f25ce9 and it passes, and a convention carries no regression to pin.
</details>

src/services/process-runner.ts
PMerlet: Closed: a `runCapture` case now fails on stdout alone and asserts both the carried message and its redaction.
<details><summary>How it was checked</summary>

The case leaves stderr empty, so the message can only come from the stdout fallback at line 401.
Ran it at 4f25ce9 and it passes, meeting `testing.md#cover-error-and-edge-paths-not-only-the-happy-path` with no pin owed.
</details>

PMerlet (approved):

